PR #9, opened 2015-10-19
Add integration test for external logger

tejohnso:
node-fetch can handle a native js object for the headers so network.js doesn't need a custom setHeaders handler.


fjvallarino:
:+1:

PR #12, opened 2015-10-19
Cleanup external logger

tejohnso:
Made a logger/bigquery directory
Moved configuration to a separate file in that dir
Made the code less java-like
Added a test to make sure token refresh isn't being called unnecessarily.


fjvallarino:
:+1:

PR #23, opened 2015-10-31
Add Windows autostart

fjvallarino:
@tejohnso please review


tejohnso:
:+1: Ok, but downloader is doing even more now.

PR #24, opened 2015-10-31
Don't user a .ver file for the installer version

tejohnso:
The installer will always know what its own version is so we shouldn't need a .ver file for that.  

This also allows the ui to show the version number for new users.


fjvallarino:
:+1:

PR #35, opened 2015-11-19
Use a temporary display id for logging new display

tejohnso:
Use a simple Math.random() "display id" on new installs.  That way the logging can keep track of log data from the same installer run.


fjvallarino:
:+1:

PR #36, opened 2015-11-20
Add automatic optimization of Windows settings

fjvallarino:
@tejohnso please review


tejohnso:
This is a case where it almost seems better to exempt the entire module from unit tests.  
I'm glad they're in there though :)
:+1:

fjvallarino:
@tejohnso Yeah, not much value added, since those are just shell commands

PR #37, opened 2015-11-20
Update os optimization

tejohnso:
@fjvallarino I refactored this quite a bit so that the commands are in an array of strings rather than a bunch of different function.  So the update process just goes through all the command strings and runs execSync on each.  

Now the test doesn't have to check for a call to many different function.  It just confirms that execSync was called for every string command.

Going through this, I noticed something I didn't realize before while studying your tests.  Node's require() returns the same object if the path points to the same file as another require().  It makes mocking easier because the require() in the test file is going to return the same object as the require() in the code under test.  So no DI necessary.  When you mock a module from a require() in the test it mocks that same module *everywhere*. Neat. I guess the pattern ```module.exports = function(dep1, dep2, dep...) {} ```isn't necessary.

### DIFF
--- a/os-optimization.js
+++ b/os-optimization.js
@@ -1,74 +1,84 @@
 var platform = require("./common/platform.js"),
-childProcess = require("child_process");
+childProcess = require("child_process"),
+linuxCommands = {
+  disableScreenSaver: [
+    "gsettings set org.gnome.desktop.session idle-delay 0"
+  ],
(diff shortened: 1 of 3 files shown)


fjvallarino:
@tejohnso Yeah, the "singleton" require is really useful. My unit tests rely on that and on mock's restore method.

:+1:

PR #38, opened 2015-11-20
Fix linux commands

tejohnso:


fjvallarino:
We'll have to modify the docs to ask users to add themselves to sudoers...

:+1:

PR #42, opened 2015-11-26
Validate CAP is not installed before running Player

fjvallarino:
@tejohnso please review


tejohnso:
We now have two versions of CAP in the wild.  We should check for **mfpgpdablffhbfofnhlpgmokokbahooi** as well.

tejohnso:
:+1:

PR #51, opened 2015-11-30
Improve watchdog check

fjvallarino:
@tejohnso please review


tejohnso:
@fjvallarino I renamed the watchdogRunning() to hasLegacyWatchdog since it wasn't only capturing running scenarios.  
Also moved those two into the prereqs file and added the checking outside of the installer into the main prereq checks and gave them messages in messages.json

fjvallarino:
:+1:

PR #52, opened 2015-12-02
Fix throw/catch

tejohnso:
Main promise chain doesn't have an error handler. Each promise handles
its own error and throws again, so the main chain executes nothing else.


fjvallarino:
:+1:

PR #56, opened 2015-12-03
Build on master and provide artifacts

tejohnso:
@fjvallarino This copies the builds to circle ci's artifacts server.  This way we don't have to upload to GCS to test out changes.


fjvallarino:
:+1:

PR #66, opened 2015-12-04
Remove Windows XP check

fjvallarino:
Remove Windows XP check. Since Electron won't start anyway, keeping the validation does not add value.


fjvallarino:
@tejohnso please review

tejohnso:
:+1:

PR #74, opened 2015-12-08
Use mkdirp to ensure paths exist for file writes

tejohnso:
@RV-Adrian was getting error on file write.  Turns out autostart and rvplayer directories weren't there so the writes failed.  This should ensure that the paths exist before trying to write the file.


fjvallarino:
:+1:

PR #79, opened 2015-12-08
Fix/update ui

RV-Adrian:
@tejohnso @fjvallarino 
Please review.
Update to the UI here. If this looks alright, I'd like to remove all inline CSS but this shouldn't hold you back from proceeding. Thanks!

### DIFF
--- a/ui/ipc.js
+++ b/ui/ipc.js
@@ -75,6 +75,8 @@ ipc.on("errorMessage", (evt, detail)=> {
   p.innerHTML = detail;
   document.querySelector("div.errors").appendChild(p);
   setContinueButtonEnabled(false);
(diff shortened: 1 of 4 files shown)


ui/ipc.js
```diff
@@ -75,6 +75,8 @@ ipc.on("errorMessage", (evt, detail)=> {
   p.innerHTML = detail;
   document.querySelector("div.errors").appendChild(p);
   setContinueButtonEnabled(false);
+  var currentSlide = document.querySelector(".container.slide.active");
+  currentSlide.className = "container slide inactive";
```
RV-Adrian: @tejohnso @fjvallarino 
Added this to remove the installation message when there is an OS incompatibility error.

tejohnso:
:+1:

PR #81, opened 2015-12-09
Handle CAP check scenario where Chrome is not installed

fjvallarino:
@tejohnso please review


tejohnso:
Poorly documented.  It should return an empty array. :)

:+1:

PR #95, opened 2015-12-12
Respect unattended mode and improve message

tejohnso:


fjvallarino:
@tejohnso Wouldn't calling sudo in unattended mode cause issues?

tejohnso:
Yes.   At the time I was imagining the sudoers would be set up correctly but that is not likely and we don't want to have to add that requirement. 

Now I think we might be able to assume that an unattended call is happening on a machine that has already had the optimisations.  In that case we can exit.   Or maybe just skip the sudo operations and perform the others.

tejohnso:
@fjvallarino please review latest changes

fjvallarino:
:+1:

PR #104, opened 2015-12-15
Show logs and errors of spawned programs

tejohnso:
@fjvallarino This shows errors from the java proxy detection.  Seems it's not able to find the main class in some cases.


fjvallarino:
:+1:

PR #107, opened 2015-12-15
Fix logging so it displays as a formatted message instead of a block of unformatted text

fjvallarino:
@tejohnso please review


tejohnso:
:+1:

PR #108, opened 2015-12-15
Stop running components before removing folders

tejohnso:
@fjvallarino This is going to fail because stopViewer doesn't exist.  

But I think we should have a general stopAll() that we can call instead of calling 4 different  methods

### DIFF
--- a/common/platform.js
+++ b/common/platform.js
@@ -37,6 +37,14 @@ module.exports = {
   getTempDir() {
     return os.tmpdir();
   },
+  getCwd() {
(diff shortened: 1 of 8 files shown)


fjvallarino:
@tejohnso please review

tejohnso:
Should we get rid of launcher.stopPlayer and launcher.stopCache since we now have a killJava

fjvallarino:
@tejohnso updated

tejohnso:
:+1:

PR #117, opened 2015-12-21
Indicate this program is compatible with Windows Vista and above to avoid Windows Program Compatibility Assistant showing up

fjvallarino:
@tejohnso please review


tejohnso:
Where does that publicKeyToken come from?

fjvallarino:
@tejohnso I'm not really sure. It comes in every example I've seen, in the automatically generated manifest and, also, in other programs whose manifest I verified.

tejohnso:
:+1:

PR #120, opened 2015-12-21
Retry network checks

tejohnso:
@fjvallarino pending some manual testing


tejohnso:
@fjvallarino please review.  Tested new build on win32 and lnx32.  Not sure if there were any network issues though.

PR #130, opened 2015-12-24
Remove "istanbul" from build's node_modules directory

fjvallarino:
Remove "istanbul" from build's node_modules directory, because it started causing installer extraction failures on Windows since late 23/12/2015

We can also consider removing other directories we don't actually need in the final artifact

@tejohnso please review


fjvallarino:
@tejohnso I'm merging this PR since I need it to make further PRs testable

PR #136, opened 2015-12-30
Chore/auto deploy

tejohnso:
@fjvallarino The remote version files were changed to json files and included in the repo so that they could be deployed to the GCS bucket.  The version of the installer version.json and version in the cfg files gets updated as part of the build.js

An interesting side-effect of having the remote components list here in the repo is that we no longer need the separate version.json file for the installer.  It can check the cfg directory to see what was deployed at the time it was built.  As other components are brought into the platform, we will be able to eliminate their .ver files as well.


fjvallarino:
:+1:

PR #138, opened 2015-12-31
Default to stable channel

tejohnso:
If urls for testing or latest are missing then default to stable


tejohnso:
@fjvallarino please review

fjvallarino:
:+1:

PR #144, opened 2016-01-04
Don't die if player is running

tejohnso:
Start player as part of launcher then check if it's running before quitting.  @fjvallarino please review


fjvallarino:
We may need to change the port until we deprecate the Java player

:+1:

PR #146, opened 2016-01-05
Use "then" method of text() response promise

fjvallarino:
@tejohnso please review


tejohnso:
:+1:    Let's see what Alan and Hannah come up with

fjvallarino:
@tejohnso I added two more commits. The one related to removing the installed versions is redundant, since we were already doing that for certain components (I'm not sure why Cache and Installer were not being removed); I think removing them unconditionally is better. The important change is the one related to creating a new temp folder for each installer run. Always using temp was the cause of having extraneous files in our install directory (since whatever we extracted previously was not removed).

tejohnso:
:+1: Let's put it through testing

fjvallarino:
@tejohnso rebased

tejohnso:
:+1:

PR #148, opened 2016-01-17
Don't allow two instances to run

tejohnso:
This is particularly important now that we have a long lived process
being spawned for the player.


fjvallarino:
:+1:

PR #149, opened 2016-01-20
Handle viewer calls

fjvallarino:

### DIFF
--- a/installer/launcher.js
+++ b/installer/launcher.js
@@ -3,26 +3,15 @@ player = require("rise-player-electron"),
 network = require("rise-common-electron").network,
 path = require("path");
 
-function getJavaPath() {
-  if(platform.isWindows()) {
-    return platform.getInstallDir() + "\\JRE\\bin\\javaw.exe";
-  }
-  else {
-    return platform.getInstallDir() + "/jre/bin/java";
-  }
-}
-
 function startCache() {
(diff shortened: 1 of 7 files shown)


tejohnso:
:+1:

installer/launcher.js
```diff
@@ -32,13 +21,12 @@ module.exports = {
     })
     .then(()=>{
       log.all("cache start", "", "50%");
-      startCache();
+      module.exports.startCache();
```
ahmedalsudani: @fjvallarino Is it necessary to call `startCache` as a child of `module.exports`?

fjvallarino:
@ahmedalsudani I did it that way so it can be mocked on unit tests; otherwise it will always call the internal version and try to start a real Java process.

ahmedalsudani:
@fjvallarino I see. Looks good then. I found 4 lint errors that have probably been there even before this PR, but do you mind fixing them?

```
installer/config.js: line 40, col 32, 'reject' is defined but never used.
installer/config.js: line 97, col 32, 'reject' is defined but never used.
installer/config.js: line 2, col 1, 'network' is defined but never used.

installer/edit-config.js: line 3, col 1, 'ws' is defined but never used.
```

Aside from that :+1:

tejohnso:
@fjvallarino I think this can be closed

fjvallarino:
@tejohnso yeah, not much is lost, and bringing it up to date would be harder than rewriting.

PR #150, opened 2016-01-20
Don't check for googleapis or catch rejections

tejohnso:
@fjvallarino This has to go in for today's release so that googleapis isn't checked.  And the catch was removed because it was catching the throw just above.  @ahmedalsudani fyi


fjvallarino:
:+1:

PR #152, opened 2016-01-21
Update versions

tejohnso:
@fjvallarino This set's the versions in the JSON files to match the currently deployed electron CFG files.
@ahmedalsudani fyi


fjvallarino:
:+1:

PR #162, opened 2016-02-04
Do not modify autostart settings if running on unattended mode

fjvallarino:


tejohnso:
:+1:

PR #166, opened 2016-02-05
Release branch

tejohnso:
@fjvallarino @ahmedalsudani Ready to merge release branch


fjvallarino:
@tejohnso @ahmedalsudani I thought we were going to provide artifacts from this branch, have CS test then and only then merge to master.

ahmedalsudani:
@fjvallarino don't we usually send master to CS for testing?

tejohnso:
@fjvallarino Master still isn't going to production so it doesn't matter at this point.  We'd have to rename this branch to get a staging build. Let's do it that way since it's more in line with master == production.

fjvallarino:
@ahmedalsudani We did that in the past, but I think we should follow the "master equals production" rule.

ahmedalsudani:
@fjvallarino Fair enough. Let's do it the right way :v:

PR #168, opened 2016-02-05
Global error catch should only log to file, not to UI (the UI part is handled by upper layers)

fjvallarino:

### DIFF
--- a/installer/installer.js
+++ b/installer/installer.js
@@ -95,10 +95,6 @@ module.exports = {
           return Promise.reject(err);
         }
       });
-    })
-    .catch((err)=>{
-      log.error(require("util").inspect(err), err.userFriendlyMessage || messages.unknown);
-      return Promise.reject(err);
     });
   },
   checkInstallerUpdateStatus() {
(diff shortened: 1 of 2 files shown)


ahmedalsudani:
:+1: :tada: let's do this

tejohnso:
How do we know for certain that all other errors do log to the UI?  If this is for unexpected errors, it seems it means it won't have been caught elsewhere.

ahmedalsudani:
:+1: on board with this too

tejohnso:
:+1:

tejohnso:
@ahmedalsudani @fjvallarino I added a commit to prevent the rest of the actions from continuing if the prereqs fail.  Haven't tested it throughly yet.

fjvallarino:
@tejohnso @ahmedalsudani This version works fine; I tested network checks, CAP and App Monitor. I also tested disconnecting Internet while components are being downloaded. In that scenario, we are showing this message: "Error downloading componentName.tar.gz. Please try again or post a question to http://community.risevision.com.". Is that correct? Do we want to tell the user which component failed or just saying we were not able to download component/update would be enough? I'm not sure we want to display that kind of detail to users (it may confusing).

tejohnso:
I think that's OK.  I think it's a sensible error and it's going to go away once we complete the next epic anyway.

PR #174, opened 2016-02-09
Fix: empty response considered yes

ahmedalsudani:
If a user is prompted to enter 'y' by initial-run, an empty
response will not cause the installation to continue

### DIFF
--- a/installer/initial-run.sh
+++ b/installer/initial-run.sh
@@ -24,7 +24,7 @@ if [ $? != 0 ]; then
   exit 0
 fi
 
-if [ $REPLY != y ]; then
+if [ "$REPLY" != y ]; then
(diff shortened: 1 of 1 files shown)


ahmedalsudani:

tejohnso:
@ahmedalsudani empty response not considered yes is the expected behaviour.  Only y is considered yes.  Empty response is not.

ahmedalsudani:
@tejohnso my mistake. I should have made that clearer. That's the new behavior in this commit.

As of this commit, if you enter an empty response, it's treated as a no.

tejohnso:
:+1:

fjvallarino:
:+1:

PR #177, opened 2016-02-10
Kill explorer.exe on Windows (except on 7) before starting Player

fjvallarino:
Kill explorer.exe on Windows (except on 7) before starting Player, to avoid issues with touchscreen right swipe bringing up Action Center

### DIFF
--- a/installer/ui/controller.js
+++ b/installer/ui/controller.js
@@ -50,5 +50,8 @@ module.exports = {
   },
   startUnattended() {
     mainWindow.webContents.send("start-unattended");
+  },
+  startUnattendedCountdown() {
+    mainWindow.webContents.send("start-unattended-countdown");
   }
 };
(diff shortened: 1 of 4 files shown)


tejohnso:
@fjvallarino This shouldn't only be for unattended from my understanding.  If a user starts the player manually they still need to prevent tampering.  

The message should use the ui/messages module and have its own .md file.  I think we're going to need a more descriptive message but we can clarify that later.

fjvallarino:
@tejohnso @ahmedalsudani updated to have explorer.exe killed on manual launcher invocations.

The message was not moved into its own Markdown module since it's part of a UI slide that may need specific styling; we are already doing something similar for the initial and proxy slides. In case you think we should change it, maybe discussing on a Hangout would be easier.

tejohnso:
@fjvallarino One more thing....can you add a --skip-timeout for manual testing scenarios as well as E2E?

fjvallarino:

tejohnso:
:+1:

PR #178, opened 2016-02-12
Set logging's main window receiver to null before closing it

fjvallarino:
Set logging's main window receiver to null before closing it, otherwise "send" messages from log.progress crash the main process on Windows

We already were on v0.36.7 which includes the fix for https://github.com/atom/electron/issues/4279

### DIFF
--- a/main.js
+++ b/main.js
@@ -63,6 +63,7 @@ app.on("ready", ()=>{
   log.debug("app ready event received");
 
   ipc.on("close", ()=>{
+    log.setUIWindow(null);
     mainWindow.close();
   });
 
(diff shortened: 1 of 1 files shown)


tejohnso:
@fjvallarino can we just use one `log.setUIWindow(null);` in the window-all-closed handler?

fjvallarino:
@tejohnso My concern in doing that is the MainWndow object could have been deleted and the event not called immediately, allowing for failures on a running download/extraction/etc background process. I also thought about adding a custom close() method to controller.js' mainWindow object, but it means having that module know about log's setMainWindow method.

tejohnso:
:+1:

ahmedalsudani:
:+1:

PR #180, opened 2016-02-15
Don't update the countdown if the window is closed

tejohnso:
@ahmedalsudani @fjvallarino I sometimes hit a window has been destroyed error when closing during countdown.


fjvallarino:
:+1:

ahmedalsudani:
:+1:

PR #181, opened 2016-02-15
Don't silently continue on error

tejohnso:
@ahmedalsudani @fjvallarino We should capture these scenarios into BQ if possible.


fjvallarino:
:+1:

ahmedalsudani:
:+1:

PR #183, opened 2016-02-16
Fix countdown message on main screen

fjvallarino:

### DIFF
--- a/installer/ui/ipc.js
+++ b/installer/ui/ipc.js
@@ -1,11 +1,17 @@
 const ipc = require("electron").ipcRenderer;
 
-ipc.on("first-ping", ()=> {
+ipc.on("first-ping", (evt, detail)=> {
   ipc.send("ui-pong");
   var closeButton = document.querySelector("#close"),
   cancelButton = document.querySelector("#cancel"),
   addToStartup = document.querySelector("#addToStartup"),
-  continueButton = document.querySelector("#continue");
+  continueButton = document.querySelector("#continue"),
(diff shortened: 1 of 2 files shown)


installer/ui/ipc.js
```diff
@@ -1,11 +1,17 @@
...
   var closeButton = document.querySelector("#close"),
   cancelButton = document.querySelector("#cancel"),
   addToStartup = document.querySelector("#addToStartup"),
-  continueButton = document.querySelector("#continue");
+  continueButton = document.querySelector("#continue"),
+  restartBlocks = document.querySelectorAll(".restartBlock"),
+  requiresReboot = detail.isWindows && detail.version !== "7";
+
+  for(var i = 0; i < restartBlocks.length; i++) {
+    restartBlocks[i].style.display = requiresReboot ? "block" : "none";
```
ahmedalsudani: @fjvallarino Should we just add a CSS rule to the page?

What do you think @tejohnso?

ahmedalsudani:
@fjvallarino aside from the nitpick above, :+1:

installer/ui/ipc.js
fjvallarino: @ahmedalsudani What do you mean? I don't think we have a way of detecting OS version using CSS (at least I have not found it).

ahmedalsudani: @fjvallarino I meant do something on the JS side like 

```js
if (win10) {
  document.styleSheets[0].addRule(...)
}
```

tejohnso:
:+1:

PR #190, opened 2016-02-22
Use a machine id for temporary display id

tejohnso:
@ahmedalsudani @fjvallarino Please review

### DIFF
--- a/installer/machine-id.js
+++ b/installer/machine-id.js
@@ -1,0 +1,24 @@
+var os = require("os"),
+crypto = require("crypto"),
+totalMem = os.totalmem(),
+interfaces = os.networkInterfaces(),
+macs = "",
+cpus;
+
+cpus = os.cpus().map((cpu)=>{
+  return cpu.model;
+}).join();
+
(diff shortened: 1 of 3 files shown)


fjvallarino:
:+1:

installer/machine-id.js
```diff
@@ -0,0 +1,24 @@
+var os = require("os"),
+crypto = require("crypto"),
+totalMem = os.totalmem(),
+interfaces = os.networkInterfaces(),
+macs = "",
+cpus;
+
+cpus = os.cpus().map((cpu)=>{
```
ahmedalsudani: Does this method always return cpus in the same order?

installer/machine-id.js
```diff
@@ -0,0 +1,24 @@
+var os = require("os"),
+crypto = require("crypto"),
+totalMem = os.totalmem(),
+interfaces = os.networkInterfaces(),
```
ahmedalsudani: @tejohnso same as the CPU question below, will we always get the interfaces in the same order?

ahmedalsudani:
:+1:

PR #195, opened 2016-02-25
Chore/use asar

tejohnso:
@fjvallarino  I added some commits here while debugging the shortcut.exe issues - please review


fjvallarino:
:+1:

PR #196, opened 2016-02-25
Continue installation on stop/start menu item err

tejohnso:
@fjvallarino I shouldn't have added a rejection for stop-start.  This will log the error but continue the install just like the uninstall.lnk failure.


fjvallarino:
@tejohnso makes sense :+1:

PR #198, opened 2016-02-26
Do not allow running installer as root

fjvallarino:
I'm sure the message could be better/clearer


tejohnso:
:+1:

PR #200, opened 2016-02-29
Don't send unknown error for extraction or removal

tejohnso:
@fjvallarino @ahmedalsudani I found another case where we're sending unknown error back unnecessarily.  Please review.


fjvallarino:
:+1:

PR #202, opened 2016-03-01
Handle error on proxy file write error

tejohnso:
@fjvallarino @ahmedalsudani write text file no longer handles errors from platform.js so this error handler is added here.


fjvallarino:
:+1:

PR #203, opened 2016-03-01
Handle uncaught errors and promises and log to BQ on next run

fjvallarino:

### DIFF
--- a/main.js
+++ b/main.js
@@ -36,6 +36,8 @@ log.external("started");
 log.debug("Electron " + process.versions.electron);
 log.debug("Chromium " + process.versions.chrome);
 
+logUncaugthErrorToBQ();
+
 app.on("window-all-closed", ()=>{
   log.debug("All windows closed");
   log.external("all closed");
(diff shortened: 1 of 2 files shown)


main.js
```diff
@@ -47,6 +49,16 @@ app.on("window-all-closed", ()=>{
 
 app.on("error", (err)=>{log.error(err, messages.unknown);});
 
+process.on("uncaughtException", (err)=>{
+  platform.writeTextFileSync(platform.getUncaughtErrorFileName(), JSON.stringify(err, ["message", "arguments", "type", "name"]));
```
tejohnso: @fjvallarino Why are we specifying fields on this one?

ahmedalsudani:
@fjvallarino once Tyler's concerns are addressed, :+1:

main.js
```diff
@@ -59,8 +71,26 @@ function isTestingMode() {
   });
 }
 
+function logUncaugthErrorToBQ() {
+  if(platform.fileExists(platform.getUncaughtErrorFileName())) {
+    try {
+      var loggedError = JSON.parse(platform.readTextFileSync(platform.getUncaughtErrorFileName()));
+
+      log.external("error", require("util").inspect(loggedError));
```
tejohnso: @fjvallarino I think we should set the main event to "uncaught error" here.

main.js
fjvallarino: @tejohnso Because stringifying an Error gives you an empty string, and if we serialize the whole object we get a stack trace which I'm not sure we really want in our logs.

fjvallarino:

tejohnso:
:+1:

PR #204, opened 2016-03-01
Use the temp displayid instead of logging null

tejohnso:

### DIFF
--- a/installer/config.js
+++ b/installer/config.js
@@ -89,13 +89,17 @@ function getDisplaySettingsFileName() {
 }
 
 function getDisplaySettingsSync() {
-  var tempDisplayId = "0." + machineId(),
+  var settings,
+  tempDisplayId = "0." + machineId(),
   textFileString = platform.readTextFileSync(getDisplaySettingsFileName());
-  if (!textFileString) {textFileString = "";}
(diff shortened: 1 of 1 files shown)


installer/config.js
```diff
@@ -89,13 +89,17 @@ function getDisplaySettingsFileName() {
...
+  var settings,
+  tempDisplayId = "0." + machineId(),
   textFileString = platform.readTextFileSync(getDisplaySettingsFileName());
-  if (!textFileString) {textFileString = "";}
-  if (textFileString.indexOf("displayid") < 0) {
-    textFileString += "\ntempdisplayid=" + tempDisplayId;
-  }
-  return parsePropertyList(textFileString);
+
+  if (!textFileString) {return {tempdisplayid: tempDisplayId};}
```
ahmedalsudani: @tejohnso `tempdisplayid` is not in camelCase, but `tempDisplayId` is. Is that intentional?

ahmedalsudani:
If the above is intentional, :+1:

installer/config.js
tejohnso: @ahmedalsudani Yeah, the format in the config file is all lowercase.

PR #205, opened 2016-03-01
Don't reject from prereq check - return false

tejohnso:
@ahmedalsudani @fjvallarino If the prereqs failed we were rejecting the promise.  That was triggering the new unhandled promise handler.   Instead the prereqs check will now return boolean pass/fail and on fail we don't run the install update or any further tasks, we just sit at whatever screen was showing (error or proxy).


fjvallarino:
:+1:

ahmedalsudani:
:+1:

PR #207, opened 2016-03-03
Show installer version and display id when an error stops the process

fjvallarino:


installer/ui/controller.js
```diff
@@ -27,7 +30,12 @@ module.exports = {
     });
 
     mainWindow.webContents.on("did-finish-load", ()=> {
-      var details = { isWindows: platform.isWindows(), version: platform.getWindowsVersion() };
+      var details = {
+        isWindows: platform.isWindows(),
+        osVersion: platform.getWindowsVersion(),
+        installerVersion: version,
+        displayId: displaySettings.displayid
```
tejohnso: @fjvallarino I think this should be displaySettings.displayid || displaySettings.tempdisplayid

fjvallarino: @tejohnso I'm handling that on ipc.js, but currently it shows a message saying the displayId has not been set.

tejohnso: @fjvallarino Showing our tempdisplayid (machine id) would be more useful than saying it's not set.

fjvallarino: @tejohnso I'll change it; I did not set it originally since I thought displaying that decimal number was a bit ugly, but it's clearly more useful

tejohnso: @fjvallarino It's ugly, but not more ugly than the fact that an error has occurred :(

The version/displayid section should be in a small font though so they're not the focus.

fjvallarino:
@tejohnso @ahmedalsudani updated. I set the font size to 11px (the rest of the message is at 12px). At 10 px it looked too small for me, but maybe it's ok for you.

Please review

tejohnso:
@fjvallarino This branch throws an error on startup if RiseDisplayNetworkII.ini does not exist.

fjvallarino:
@tejohnso where? config.getDisplaySettingsSync does not throw when the file does not exist (it relies on platform.readTextFileSync)

installer/ui/controller.js
```diff
@@ -27,7 +30,12 @@ module.exports = {
     });
 
     mainWindow.webContents.on("did-finish-load", ()=> {
-      var details = { isWindows: platform.isWindows(), version: platform.getWindowsVersion() };
+      var details = {
+        isWindows: platform.isWindows(),
+        osVersion: platform.getWindowsVersion(),
```
ahmedalsudani: @fjvallarino so this will change to `platform.getOsVersion()` if we also add Linux?

ahmedalsudani:
@fjvallarino just wondering re: the question above. Once Tyler's concerns are addressed, :+1:

installer/ui/controller.js
fjvallarino: @ahmedalsudani well, getWindowsVersion has some logic for providing a human readable Windows version number, so replacing with getOsVersion will not work. In that case we'll have to rename this property.

tejohnso:
@fjvallarino platform.readTextFileSync() doesn't throw either when there's no text file, but it does try to log.  And this is all happening before the logger exists.  It might be better to pass the display settings into the controller similar to how they're passed into the logger in line 30 in main.js

fjvallarino:

tejohnso:
@fjvallarino I changed the order around so that the displayid and version info is at the bottom, and the more important part (the message) is at the top.  And I moved the id/version into one line.  Just to try to further minimize their focus. Ready for merge if you are.
@ahmedalsudani fyi

fjvallarino:
@tejohnso ok :+1:

PR #209, opened 2016-03-04
Remove previous installer folder before upgrading

fjvallarino:


ahmedalsudani:
:+1:

installer/installer.js
```diff
@@ -127,7 +127,10 @@ module.exports = {
   updateInstaller(installerPkgTempPath) {
     var targetDir = path.join(platform.getInstallDir(), config.getComponentInfo("InstallerElectron").copy);
 
-    return platform.copyFolderRecursive(installerPkgTempPath, targetDir)
+    return platform.deleteRecursively(targetDir)
```
tejohnso: @fjvallarino I think you should put a catch here so that if the deletion fails half way through it still goes ahead with the copy.

fjvallarino: @tejohnso if we can't remove the folder it's highly probable we will encounter issues while copying (because of overwriting), and we will leave the folder in an inconsistent state (similar to what's currently causing the issue).

tejohnso: @fjvallarino In the case that we don't remove anything, I agree it's best not to bother with the copy.  But what if we end up partially removing the folder?  Then it's definitely a broken player.  And a copy attempt should be made.  I think it's safer to make sure we avoid the definitely broken scenario.  Any other options?

fjvallarino: @tejohnso until we move to the multiple versions scenario, we'll always have potential delete/overwrite issues. I'll add the catch clause.

fjvallarino:
@tejohnso @ahmedalsudani updated to ignore removeFolder failures and attempt to copy new installer

tejohnso:
:+1:

PR #212, opened 2016-03-07
Add graceful shutdown for RisePlayer on Windows

ahmedalsudani:


fjvallarino:
:+1:

tejohnso:
@ahmedalsudani Why don't we use powershell to place the graceful shutdown file and then taskkill?  Seems simpler than going through the network.

ahmedalsudani:
@tejohnso how do we know that player is running when we issue taskkill?

tejohnso:
@ahmedalsudani good point.  And what about including the PS1 file as part of the repo instead constructing it as part of the install process?

ahmedalsudani:
@tejohnso that we can do. Is there a reason we don't include the batch files in the repo?

tejohnso:
@ahmedalsudani No reason I can think of.

fjvallarino:
@ahmedalsudani @tejohnso the main reason is the scripts use paths which depend on the user's folder or on OS specific folders which may not always be in the same location. We can use the corresponding environment variables, I guess.

ahmedalsudani:
@tejohnso I experimented with having the scripts in separate files, but it does not always work. Specifically, it doesn't work when we're calling the `stop-player.ps1` script. The path to the script cannot be relative, so we have to use `platform.getInstallDir()` in order to call it.

I could move out `stop-player.ps1` alone. However, I don't see what is gained by moving one script into a separate file and leaving 3.

ahmedalsudani:
@tejohnso are you ok with this PR?

tejohnso:
@ahmedalsudani The path is always "%LOCALAPPDATA%/rvplayer/Installer" on windows and "$HOME/rvplayer/installer" on linux.

ahmedalsudani:
@fjvallarino @tejohnso code is ready for review. Before merging, I need to 
- rebase
- update tests

I chose not to move the uninstallation script out of the js file because it involves deleting the shortcuts, which is the most complicated part. If you don't think it would be a problem, let me know

build.js
```diff
@@ -130,6 +131,7 @@ function createSelfExtractingInstallers() {
     "RunProgram=\"installer.exe\"" + "\n" +
     ";!@InstallEnd@!" + "\n";
 
+    spawnSync("sh", ["-c", "cp -r scripts builds/installer-" + platform]);
```
tejohnso: @ahmedalsudani each argument should be a separate element in the array.   "cp -r .." should be all together.

ahmedalsudani: @tejohnso isn't that the case now?

tejohnso: Yeah, looks like it is.

tejohnso:
@ahmedalsudani This pr should be closed and a new one created against release-branch-7

PR #213, opened 2016-03-08
Chore/components removal

tejohnso:
@ahmedalsudani @fjvallarino Here's what I have so far on the component removal update for the installer.  Still very much a work in progress.


tejohnso:
@ahmedalsudani I was thinking we'd always move the version forward similar to a git revert making a new commit.  Maybe that's too complicated though, because we'd have to also release another build as part of the rollback.

PR #215, opened 2016-03-09
Disable pinch-to-zoom in Installer

ahmedalsudani:
@fjvallarino @tejohnso this fixes https://github.com/Rise-Vision/rise-player-electron/issues/14#issuecomment-194393752. It only covers the installer, not presentations running in Chrome.


fjvallarino:
:+1:

installer/ui/ipc.js
```diff
@@ -1,4 +1,8 @@
 const ipc = require("electron").ipcRenderer;
+const webFrame = require("web-frame");
```
tejohnso: The preferred format for this is now require('electron').webFrame as in ipcRenderer above.

ahmedalsudani:
@tejohnso updated

tejohnso:
:+1:

PR #216, opened 2016-03-10
Chore/fix spawn

tejohnso:
@ahmedalsudani @fjvallarino please review.  Fixes the proxy issue, prevents the proxy screen from showing up repeatedly, and adds a rollout percentage override.


installer/component.js
```diff
@@ -194,6 +203,7 @@ module.exports = {
   getTestingVersion,
   isPlayerOnLatestChannelVersion,
   getChannel,
+  overriddenPct() { return overriddenPct; },
```
ahmedalsudani: @tejohnso should this be `getOverriddenPct()` to make it clear it's a function?

installer/component.js
```diff
@@ -48,7 +50,14 @@ function getTestingVersion() {
 }
 
 function getChannel(components) {
-  percentageIsMet = module.exports.getLatestChannelProb() < components.LatestRolloutPercent;
+  overriddenPct = yargs.parse(process.argv.slice(1)).rolloutPct;
+  log.debug(`Using rollout percentage: ${overriddenPct}`);
+
+  if (overriddenPct || overriddenPct === 0) {
```
ahmedalsudani: @tejohnso we can use `overriddenPct != null` here

installer/network-check.js
```diff
@@ -39,18 +39,30 @@ module.exports = {
...
+    var command = platform.getJavaExecutablePath(),
+    args = Array.from(network.getJavaProxyArgs());
 
-    retryCount = retryCount === undefined ? 5 : retryCount;
+    args.push("-jar");
+    args.push(path.join(platform.getInstallerDir(), "java-network-test.jar"));
+    args = args.concat(siteList);
 
-    return platform.spawn(command)
+    retryCount = retryCount === undefined ? 2 : retryCount;
```
ahmedalsudani: @tejohnso why are we reducing the default retry count?

tejohnso: @ahmedalsudani It was mostly pointless after the third try.

installer/component.js
tejohnso: @ahmedalsudani The syntax doesn't lie :)

fjvallarino:
:+1:

main.js
```diff
@@ -106,16 +106,18 @@ app.on("ready", ()=>{
...
       if (!passed) {return;}
 
       return installer.begin()
-      .then(postInstall)
+      .then(()=>{
+        return postInstall({withProxy: false});
+      })
       .then((passed)=>{
-        if (!passed) {return;}
+        if (passed) {ui.enableContinue();}
```
ahmedalsudani: @tejohnso not sure if returning from a promise makes a difference, but this function stood out because it doesn't have a return statement.

tejohnso: Here it doesn't make a difference because there's no further chaining.

main.js
```diff
@@ -236,11 +244,16 @@ app.on("ready", ()=>{
     .catch(()=>{return false;});
   }
 
-  function prereqCheck() {
-    return platform.onFirstRun(()=>{
-      log.all("checking network requirements", "", "25%");
-      return prereqs.checkNetworkConnectivity();
-    })()
+  function prereqCheck(options) {
```
ahmedalsudani: @tejohnso when is `prereqCheck` called with `options = {withProxy: true}`?

tejohnso: True is the default.  It's in the main installation flow on line 140

main.js
ahmedalsudani: Ok, I still think it makes sense to return at the end if that's necessary for chaining. It's promise etiquette ;)

ahmedalsudani:
@tejohnso once you decide what to do about https://github.com/Rise-Vision/rise-launcher-electron/pull/216#discussion_r55746286, https://github.com/Rise-Vision/rise-launcher-electron/pull/216#discussion_r55751147

:+1:

installer/component.js
tejohnso: I'd rather not rely on the coercion.  I might look at that later and think that I can then change `if (overridenPct != null)` to `if (overriddenPct)`

main.js
tejohnso: Agreed.  And it's one less foot gun in the barn.

PR #220, opened 2016-03-15
Do not set logger.uiWindow to null

ahmedalsudani:

- should keep the window updated with messages in the event
  of failures
- might fix unandled exceptions thrown by electron on shutdown
- destroyed window will be handled by logger


fjvallarino:
@ahmedalsudani this was added because of the "[Object deleted]" popups when closing the main window. How are we going to handle those?

ahmedalsudani:
@fjvallarino yeah, but it was causing other problems (the JS mainly stacktraces that print on exit).

It's now being checked for [here](https://github.com/Rise-Vision/rise-common-electron/pull/31)

PR #224, opened 2016-03-15
Do not set logger.uiWindow to null

ahmedalsudani:
- should keep the window updated with messages in the event
  of failures
- might fix unandled exceptions thrown by electron on shutdown
- destroyed window will be handled by logger


ahmedalsudani:
@fjvallarino @tejohnso continued from #220

tejohnso:
:+1:

fjvallarino:
:+1:

PR #226, opened 2016-03-18
Check for valid mainWindow before calling mainWindow.send

ahmedalsudani:


installer/ui/controller.js
```diff
@@ -4,6 +4,14 @@ shell = require("electron").shell,
...
 mainWindow;
 
+function ifValidMainWindow(fn) {
+  return () => {
+    if (mainWindow && !mainWindow.isDestroyed()) {
+      return fn.apply(null, arguments);
+    }
+  };
+}
+
```
tejohnso: @ahmedalsudani why did you go with this over something like

```javascript
function validateWindow() {return mainWindow && ....}
...
 validateWindow() && mainWindow.webContents.send("message")
```

which I think is what you had before?

ahmedalsudani: @tejohnso 

in `logger.js`, we already had if clauses, so I replaced the condition with `validWindow()`. In `controller.js`, we don't have an if clause for every call, so I created a wrapper function instead. Advantages:

- cleaner
- code reuse

tejohnso:
:+1:

PR #228, opened 2016-03-18
Build self contained bundle

fjvallarino:


tejohnso:
:+1:

build-components-packages.sh
```diff
@@ -0,0 +1,72 @@
+CHROME_VERSION=48.0.2564.116
```
ahmedalsudani: @fjvallarino will we have to manually maintain the versions here?

fjvallarino: @ahmedalsudani we will have to manually insert the .ver files until we get rid of them. This script is not actually important, it's something I created to build the new tar.gz we use in the build process from the tar.gz files we use in the current installer.

installer/initial_run.bat
```diff
@@ -0,0 +1,9 @@
+@echo off
+
+md %LOCALAPPDATA%\rvplayer\VERSION
+
+echo Copying files
+robocopy /E /NFL /NDL /NJH /NJS /NC /NS /NP VERSION %LOCALAPPDATA%\rvplayer\VERSION
```
ahmedalsudani: :laughing: `robocopy`

ahmedalsudani:
:+1:

PR #244, opened 2016-03-22
Handle error type as text to avoid issues with Exceptions' properties not being enumerable

fjvallarino:


tejohnso:
:+1:

ahmedalsudani:
:+1: 

@fjvallarino was there an issue with the previous approach?

fjvallarino:
@ahmedalsudani yeah, Errors do not have enumerable properties, so their content was being dropped (only "Exception" or "Rejection" was being logged)

ahmedalsudani:
I see

PR #247, opened 2016-03-23
Log beginning install every time we check for an upgrade

ahmedalsudani:
@fjvallarino @tejohnso this should fix the issue where our logs were going > 100%


fjvallarino:
:+1:

tejohnso:
:+1:

PR #248, opened 2016-03-23
Don't abandon launch on java kill failure

tejohnso:
@ahmedalsudani @fjvallarino please review.  If all retries fail, last resort is to try launching anyway.


fjvallarino:
:+1:

ahmedalsudani:
:+1:

PR #261, opened 2016-03-24
Hide Command Prompt Windows

ahmedalsudani:
This should allows us to run commands in the background

@tejohnso this is the script. I've confirmed it passes parameters to the command it calls, and it works even if the username contains a space.

I'm working on changing all of our calls to use `background.jse` now.


tejohnso:
:+1:

ahmedalsudani:
@tejohnso updated the shortcuts and `initial-run.bat`. I'm testing one additional thing: after copying the files in `initial-run.bat`, close the command prompt window once we launch installer. I'll push that change if it works

tejohnso:
:+1:

PR #265, opened 2016-03-29
Extract bundle inside rvplayer

fjvallarino:
Windows 7z SFX version needed to be replaced since the default one does not support extraction into a specific folder.

Since CircleCI is still using Ubuntu 12.04 as its default image, makeself is largely outdated and its --target parameter is not supported; that's the reason it's currently being downloaded from the GitHub repository. Apparently, a new image should be available soon; I modified the CircleCI build script to handle this update instead of reflecting that workaround in build.sh


ahmedalsudani:
@fjvallarino much better!

:+1:

tejohnso:
:+1:

PR #267, opened 2016-03-29
Start Download after launching Player

ahmedalsudani:


main.js
```diff
@@ -151,7 +151,7 @@ app.on("ready", ()=>{
   ipc.on("install", ()=>{
     ui.disableContinue();
 
-    prereqCheck()
+    prereqCheck()()
```
ahmedalsudani: @fjvallarino @tejohnso how do you feel about this?

tejohnso: @ahmedalsudani Did you do that just so that the promise chaining is cleaner in line 186 below?

ahmedalsudani: @tejohnso I had to make this change not for cleanliness, but because `prereqCheck` is not the first function in the promise chain.

So if we want to call `prereqCheck` with parameter `{withProxy: false}`, `prereqCheck({withProxy: false})` needs to return a function that we can use in `.then(prereqCheck({withProxy: false}))`.

Because `prereqCheck()` returns a function, we need to call it then call the function it returns.

ahmedalsudani:
@tejohnso updated

@fjvallarino FYI

tejohnso:
:+1:

PR #271, opened 2016-03-31
Report free disk space

fjvallarino:

The build is failing because rise-common-electron changes are not merged; I will rebuild as soon as they are merged.


installer/installer.js
```diff
@@ -21,6 +21,14 @@ module.exports = {
     })
     .catch((err)=>{
       log.external("upgrade failure", require("util").inspect(err));
+
+      return platform.getFreeDiskSpace()
+      .then((space)=>{
+        log.external("disk space", space);
```
tejohnso: @fjvallarino This should catch disk space problems on download, but I think we should also add a log in *main* after `showUpgradeDelay` so that we catch the case where the extraction began but failed due to disk space.  If after some time the bundle hasn't started and closed this installer then it might be a disk space issue.

fjvallarino:
Updated handling of tryUpdate result

main.js
```diff
@@ -122,14 +139,21 @@ app.on("ready", ()=>{
...
-          showUpgradeDelay();
+      .then((result)=>{
+        var promise = Promise.resolve();
 
+        if (result.bundleRan) {
+          scheduleDiskSpaceCheck();
+          showUpgradeDelay();
           return false;
         }
+        else if (result.downloaded) {
```
tejohnso: @fjvallarino Don't we want to checkAvailableDiskSpace if it **wasn't** downloaded?

fjvallarino: @tejohnso probably the name of the field is not the best one; the previous check (result.bundleRan) means we were able to download and run the bundle. The second one means the bundle failed to run for some reason, although it was downloaded.

The download field could be false because there is not need for an upgrade, or because the request failed (which would not imply disk space issues). I'll rename the fields to be clearer about the intention and handle one case I had missed.

fjvallarino:

tejohnso:
:+1:

PR #274, opened 2016-04-01
Remove previous versions using build time provided scripts

fjvallarino:
Remove previous versions using build time provided scripts (in the future the scripts could be automatically generated)


ahmedalsudani:
:+1:

@fjvallarino Will the temporary files be addressed now or in another sprint?

main.js
```diff
@@ -192,6 +193,7 @@ app.on("ready", ()=>{
         })
         .then(platform.killExplorer)
         .then(launcher.launch)
+        .then(cleanup.removePreviousVersions)
```
tejohnso: I think we should move this into postInstall after setAutostart.  And only after a successful set.  It's at that point we've officially moved to the new version and expect everything to be ok with it.  If the autostart setting fails, we're not really on the new version yet and we shouldn't delete old versions.  @ahmedalsudani @fjvallarino

fjvallarino:

@ahmedalsudani we'll have to come up with a way of automating this; the bucket is not public, so listing its contents is not possible. Maybe we can append a new version number to a "versions" file during build on master (and upload it to GCS).

@tejohnso autostart is the only link creation step which rejects on failure, so adding it after that step was good enough.

fjvallarino:
@tejohnso @ahmedalsudani for Windows a "script" like this will be needed for cleaning old temp directories (it runs fine as part of a .bat file):

for /d %G in ("%LOCALAPPDATA%\Temp\rvplayer*") do rd /s /q "%~G"

Let me know if it's ok for you

tejohnso:
That script option seems OK by me.

fjvallarino:

tejohnso:
:+1:

PR #283, opened 2016-04-08
Read version before initializing logger

ahmedalsudani:
@fjvallarino @tejohnso we were getting null for version number on BQ


tejohnso:
:+1:

fjvallarino:
:+1:

PR #285, opened 2016-04-08
Use %%G instead of %G (%G is for command line, %%G for batch files)

fjvallarino:


tejohnso:
:+1:

PR #287, opened 2016-04-11
Report disk space on unattended mode

fjvallarino:
This PR includes the fix for logging disk space when an unattended upgrade fails. The main window is not closed on upgrade, since it should be closed automatically by the new installer instance; in case it's not killed, it will be closed after logging disk space and a small delay. The "running bundle" progress bar is also displayed.

Currently logging to BQ is failing (local logging works), and will be fixed in the next release.

@ahmedalsudani please review

@tejohnso FYI


installer/main-controller.js
```diff
@@ -192,11 +195,13 @@ module.exports = {
             .catch((err)=>{
               log.external("error", require("util").inspect(err));
               return;
+            })
+            .then(()=>{
+              if(!bundleLaunched) {
+                mainWindow.close();
```
ahmedalsudani: @fjvallarino is it possible to promisify the function above (`scheduleDispSpaceCheck`) so that we can do something like the following?

```js
module.exports.showUpgradeDelay()
.then(scheduleDiskSpaceCheck)
.then(mainWindow.close)
```

Then we wouldn't need to call `mainWindow.close()` inside `scheduleDispSpaceCheck`

fjvallarino: @ahmedalsudani the issue is we don't want to close the window in case an upgrade needs to be processed, because on main window close the process exits and cancels the timer.

ahmedalsudani: @fjvallarino that makes sense. I'm just wondering whether we can simplify it by putting everything in a promise chain. We can have `scheduleDiskSpaceCheck` not resolve until the timeout elapses. 

Then it won't be necessary to have two calls to `mainWindow.close()`. We can only have one call that we put at the end of the promise chain, and we know we'll only get to it after the "disk space check" delay is over. Does this make sense or is there a problem that I'm missing?

fjvallarino: @ahmedalsudani it could be harder than that, since `scheduleDiskSpaceCheck` is not the only function we're calling in that section (it's only invoked if the upgrade succeeded, otherwise the disk space check is done immediately). 

If you can think of a clean implementation for this, we can add it for release 9 (we have one card for fixing the logging issue, so we can do it in that same place).

ahmedalsudani:
Sounds good @fjvallarino 

:+1:

PR #299, opened 2016-04-15
Chore/electron viewer

ahmedalsudani:
@fjvallarino @tejohnso


viewer/controller.js
```diff
@@ -0,0 +1,78 @@
...
+  const params = {
+    type: "display",
+    "id": displaySettings.displayid || "",
+    "claimId": displaySettings.claimid || "",
+    "os": (platform.isWindows() ? "win" : "lnx"),
+    "jv": readVerFile("java"), // Java version
+    "pn": "RisePlayer", // Player name
+    "pv": readVerFile("RisePlayer"), // Player version
+    "ev": readVerFile("RiseCache"), // Cache version
+    "ip": null
```
ahmedalsudani: @fjvallarino @tejohnso I got those from https://github.com/Rise-Vision/rise-player/blob/88d897913958ab5442e4e4e5dcaf72bb0cab28bd/src/main/java/com/risevision/riseplayer/utils/SystemInfo.java#L31

They are mostly intiialized in config. For example: https://github.com/Rise-Vision/rise-player/blob/88d897913958ab5442e4e4e5dcaf72bb0cab28bd/src/main/java/com/risevision/riseplayer/Config.java#L91

PR #310, opened 2016-04-21
Don't change autostart unless user has requested it

tejohnso:
If we're in unattended mode the autostart file must be updated in case
the version changed on upgrade.  But that should only happen if the
autostart file exists.  If it doesn't exist we shouldn't create it.

Sometimes we're restarting the system in unattended mode (eg from
viewer) and we don't want autostart set if the user hasn't indicated so
during install.


installer/autostart/autostart.js
```diff
@@ -1,38 +1,47 @@
...
 
 module.exports = {
   requested(yesOrNo) {userWantsAutostart = yesOrNo;},
   setAutostart() {
-    var ubuntuAutostartPath = path.join(platform.getAutoStartupPath(), "rvplayer.desktop");
-    var windowsShortCutPath = path.join(platform.getAutoStartupPath(), "Rise Vision Player.lnk");
     var oldShortCutPath = path.join(platform.getAutoStartupPath(), "Start Rise Vision Player.lnk");
 
-    if (!userWantsAutostart) {
+    if (userWantsAutostart === false) {
```
ahmedalsudani: @tejohnso what if it's undefined?

installer/autostart/autostart.js
```diff
@@ -1,38 +1,47 @@
...
       log.debug("Removing autostart");
 
       if (platform.isWindows()) {
-        return platform.deleteRecursively(windowsShortCutPath)
+        return platform.deleteRecursively(module.exports.autostartPath())
```
ahmedalsudani: @tejohnso we can get rid of this call and ... (see below)

installer/autostart/autostart.js
```diff
@@ -1,38 +1,47 @@
...
         .then(platform.deleteRecursively.bind(null, oldShortCutPath));
       } else {
```
ahmedalsudani: in conjunction with the above, get rid of the else here

installer/autostart/autostart.js
tejohnso: @ahmedalsudani I kept it that way because of the promise chaining.  But I think we can call the deletion on the old shortcut path if it's windows and ignore that returned promise.  Then return only the result of deleting the autostartpath.

ahmedalsudani:
:+1:

PR #312, opened 2016-04-21
If run as root, set rvplayer owner as user

ahmedalsudani:
@fjvallarino @tejohnso simple enough :)


tejohnso:
@ahmedalsudani does `logname` work?  On my (non ubuntu) machine it says "no login name"

And does it work if the user is logged in as root?

fjvallarino:
@tejohnso @ahmedalsudani when running a command with sudo, the $HOME variable still points to the user's home directory, and $(logname) points to the correct user name. I don't think we should care about logging as root, since it has to be explicitely enabled on Ubuntu to be able to login with that user.

ahmedalsudani:
@tejohnso I misread your comment and thought you were talking about an Ubuntu machine. `logname` does work on Ubuntu according to my tests

tejohnso:
:+1:

fjvallarino:
:+1:

PR #317, opened 2016-04-27
Improve cleanup of previous versions and temp directory

fjvallarino:


tejohnso:
:+1:

scripts/remove-previous.sh
```diff
@@ -2,31 +2,5 @@
...
-rm -rf "$HOME/rvplayer/2016.03.29.21.17"
-rm -rf "$HOME/rvplayer/2016.03.30.12.26"
-rm -rf "$HOME/rvplayer/2016.03.30.14.36"
-rm -rf "$HOME/rvplayer/2016.03.31.03.23"
-rm -rf "$HOME/rvplayer/2016.03.31.15.00"
-rm -rf "$HOME/rvplayer/2016.03.31.15.23"
-rm -rf "$HOME/rvplayer/2016.03.31.16.01"
-rm -rf "$HOME/rvplayer/2016.03.31.16.12"
+# Old installer directories
+ls "$HOME/rvplayer" | grep "20" | grep -v VERSION | xargs -d"\n" rm -rf
```
ahmedalsudani: @fjvallarino don't we need to add `"$HOME/rvplayer"` in front of the directory names passed to `rm -rf`?

So we'd have `... | xargs -d"\n" rm -rf "$HOME/rvplayer/{}"`

tejohnso: Also we shouldn't grep on ls output
Something like `find $HOME/rvplayer -maxdepth 1 -type d -path "*/rvplayer/20` would be better
`

fjvallarino: @tejohnso what's the advantage of doing that?

fjvallarino:
@tejohnso @ahmedalsudani updated to use find instead of ls

scripts/remove-previous.sh
```diff
@@ -2,31 +2,5 @@
...
-rm -rf "$HOME/rvplayer/2016.03.29.21.17"
-rm -rf "$HOME/rvplayer/2016.03.30.12.26"
-rm -rf "$HOME/rvplayer/2016.03.30.14.36"
-rm -rf "$HOME/rvplayer/2016.03.31.03.23"
-rm -rf "$HOME/rvplayer/2016.03.31.15.00"
-rm -rf "$HOME/rvplayer/2016.03.31.15.23"
-rm -rf "$HOME/rvplayer/2016.03.31.16.01"
-rm -rf "$HOME/rvplayer/2016.03.31.16.12"
+# Old installer directories
+find "$HOME/rvplayer" -maxdepth 1 -type d -path "*20*" | grep -v VERSION | xargs -d"\n" rm -rf
```
tejohnso: I think we need to search for  "/rvplayer/*20*" 
We don't want to delete dirs like /my/home/dir/project20/stuff

ahmedalsudani:
@fjvallarino once @tejohnso's last comment is addressed
:+1:

scripts/remove-previous.sh
fjvallarino: @tejohnso it won't, since find is looking for directories inside "$HOME/rvplayer" and -path is applied later

fjvallarino:
@tejohnso @ahmedalsudani added extra safety measure

tejohnso:
:+1:

PR #324, opened 2016-04-29
Fix/explorer launch

tejohnso:
@ahmedalsudani please review as discussed.  Quit on windows calls background.jse through cmd /c - otherwise spawn throws unknown error.


ahmedalsudani:
:+1:

PR #328, opened 2016-05-03
Parallel E2E

ahmedalsudani:
@fjvallarino @tejohnso unfortunately, parallel builds did not speed up builds in any significant way on CircleCI.

This PR only makes E2E tests run in parallel and changes the E2E test output


tejohnso:
:+1:  This one is a good candidate for squashing

PR #334, opened 2016-05-06
Implement error screen

tejohnso:
@ahmedalsudani @fjvallarino please review.   I've left some debugging code in here until merge into R12.


installer/ui/js/controllers/error-screen-controller.js
```diff
@@ -0,0 +1,12 @@
...
+.controller("ErrorScreenCtrl", ["$scope", "$sce", "SharedState",
+function($scope, $sce, sharedState) {
+  $scope.errorMessage = $sce.trustAsHtml(sharedState.getState().errorMessage);
+
+  sharedState.registerObserver((state)=>{
+    console.log("error msg is " + state.errorMessage);
+    $scope.errorMessage = $sce.trustAsHtml(state.errorMessage);
+  }, "ErrorScreenCtrl");
+
+  $scope.exit = ()=>{window.close();};
```
ahmedalsudani: @tejohnso is this supposed to call `MainProcess.close`, or is `window.close` defined somewhere?

installer/ui/js/services/shared-state.js
```diff
@@ -1,22 +1,30 @@
...
 
-      set (name, val) {
-        state[name] = val;
-        // Notify observers
+      set(stateChange) {
+        state = Object.assign(state, stateChange);
+        console.log(state);
+        console.log(`calling ${observers.length} observers`);
         observers.forEach((observer)=>observer(state));
+        $rootScope.$digest();
```
ahmedalsudani: @tejohnso this might cause issues if a digest loop is already taking place.

There are two things we can do about this:

- replace this line with `setTimeout($rootScope.$digest)` -- easier
- add `$rootScope.$digest()` to all event handlers that cause state change

installer/ui/js/controllers/error-screen-controller.js
tejohnso: @ahmedalsudani It's defined in the dom

ahmedalsudani:
:+1:

PR #336, opened 2016-05-09
Implement Miranda Intro Screen

ahmedalsudani:
@fjvallarino @tejohnso ready for review.

Tests are being fixed


installer/main-controller.js
```diff
@@ -137,8 +138,9 @@ module.exports = {
       });
   },
 
-  install() {
+  install(displayId) {
     ui.disableContinue();
+    module.exports.writeConfigurationFile(displayId);
```
tejohnso: @ahmedalsudani Isn't this going to overwrite the existing config file?  What if there was something other than a displayId in there?

ahmedalsudani: @tejohnso yeah, it would. The assumption I made is that if you're running V3, the only thing in the configuration file is the display ID.

tejohnso: We also have the claim id in there.  And the viewer and core urls.  I don't think it's a good idea to overwrite the file and loose any other data.  At very least we should maintain the claim id so that subsequent runs would show it.

ahmedalsudani: Ok, going to update once I return

tejohnso:
@ahmedalsudani I can't get a displayId to validate through the UI.  I tried WXE34FGSC9NW in line 8 of the integration test and it worked but in the UI it failed.

installer/ui/js/controllers/setup.js
```diff
@@ -2,12 +2,23 @@ angular.module('riseInstaller')
...
     this.userInput = {
       type: "display", // Whether it's Display ID or  Claim ID
-      displayId: "",
+      displayId: null,
       claimId: "",
       name: ""
     };
 
+    SharedState.registerObserver((newState)=>{
+      if (!this.userInput.displayId && !/^0./.test(newState.ipcConstants.displayId)) {
```
tejohnso: @ahmedalsudani This is causing the UI to make it impossible to clear out the displayID field if there's a displayid entry in the config file.  As soon as you clear it out, userInput.displayId is falsy, and it resets it to whatever is in the file.

tejohnso:
:+1:

PR #338, opened 2016-05-10
Add Complete screen event handling

fjvallarino:


installer/ui/js/services/complete-process.js
```diff
@@ -0,0 +1,6 @@
+angular.module('riseInstaller')
+  .service('CompleteProcess', function() {
+    return {
+      startPlayer: ()=>ipc.send("launch")
+    };
+  });
```
tejohnso: @fjvallarino all of our IPC calls are currently in the MainProcess service.  
It might be better to do it this way, but I think we should choose one way.
Let's talk about it after standup.

installer/ui/partials/installation-complete.html
```diff
@@ -1,20 +1,18 @@
...
-    <button class="btn btn-primary">Start Rise Player</button>
-    <br/>
-    <br/>
-    <div class="text-center alert alert-info"> <b>TIP:</b> Press <b>CTRL-Q</b> to stop Rise Player once started.</div>
+  </div>
+  <button ng-click="vm.close()" class="btn btn-default">Exit</button>
+  <button ng-click="vm.startPlayer()" class="btn btn-primary">Start Rise Player</button>
+  <br/>
+  <br/>
+  <div class="text-center alert alert-info"> <b>TIP:</b> Press <b>CTRL-Q</b> to stop Rise Player once started.</div>
```
ahmedalsudani: @fjvallarino instead of Ctrl+Q, this should use the `stopKey` member of shared state. See here for how it's implemented in my PR https://github.com/Rise-Vision/rise-launcher-electron/pull/336/files#diff-ed7d75191d5d59e1edae6b9ab9fcfc4dR12

fjvallarino:
@tejohnso @ahmedalsudani updated with the correct key bindings for exiting the player

ahmedalsudani:
:+1: 

In the future, I think we should use the structure used in this PR (controller has service associated with it that does the work).

tejohnso:
:+1:

PR #343, opened 2016-05-10
Stop using enable / disable continue button

tejohnso:


ahmedalsudani:
@tejohnso are we going to disable the continue button some other way?

tejohnso:
@ahmedalsudani We definitely won't be doing something like `document.querySelector("#continue");`

I think it'll be conditional upon some other state that a particular controller is looking for.

ahmedalsudani:
:+1:

PR #350, opened 2016-05-12
Verify proxy at startup and add explicit skip

tejohnso:
@fjvallarino @ahmedalsudani  Please review


installer/config.js
```diff
@@ -1,15 +1,25 @@
...
 machineId = requireRoot("installer/machine-id.js"),
 electronAppPath = "",
-path = require("path");
+path = require("path"),
+proxyConfig,
+uiWindow;
 
-proxy.observe(saveProxySettings);
-function saveProxySettings(fields) {
+proxy.observe(persistProxySettings);
```
ahmedalsudani: Where is `persistProxySettings` defined?

installer/ui/js/controllers/proxy.js
```diff
@@ -1,15 +1,33 @@
...
+    this.proxySetup.validating = true;
+    this.proxySetup.errorMessage = " ";
+    ProxyService.validateProxy({
+      hostname: this.proxySetup.hostname,
+      port: this.proxySetup.port
+    });
+  };
+
+  ProxyService.registerListener((details)=>{
+    this.proxySetup = Object.assign({}, this.proxySetup, details, {validating: false});
```
ahmedalsudani: Why do we always set `proxySetup.validating` to `false` here?

installer/ui/partials/intro.html
```diff
@@ -13,8 +13,9 @@ <h1 id="playerInstallationHeader" class="half-bottom">Install Rise Player</h1>
 
   </div>
 
-  <div class="main-content">
-    <ng-include src="'partials/setup-form.html'" scope="" onload=""></ng-include>
+  <ng-include src="'partials/network-check-overlay.html'" ng-show="vm.validatingProxy"></ng-include>
+  <div class="main-content" ng-show="vm.validatingProxy === false">
```
ahmedalsudani: instead of ng-show, we can use ng-if (https://code.angularjs.org/1.1.5/docs/api/ng.directive:ngIf) which would do more than just hiding the partial.

ahmedalsudani: @tejohnso replacing the ng-show directive with `ng-show="vm.validatingProxy"` made the UI visible again on Windows. Not sure why it wasn't  being set to false.

ahmedalsudani:
Once my latest comment is addressed, we can merge and fix any issues in `feature/miranda` branch

PR #351, opened 2016-05-12
Set validating to false even if unsuccessful

tejohnso:
@ahmedalsudani @fjvallarino please review.  This was causing the "checking network" message to show on the intro tab even if the proxy screen was skipped via **continue**


fjvallarino:
:+1:

installer/config.js
```diff
@@ -44,13 +44,14 @@ function verifyProxyConfig(config) {
   .then(networkCheck.checkSitesWithJava)
   .then(()=>{
     uiWindow.send("proxy-validation-result", {validated: true});
-    uiWindow.send("validating-proxy-config", false);
```
ahmedalsudani: @tejohnso we don't set it to false if it succeeds?

PR #352, opened 2016-05-12
Set validating to false even if unsuccessful - add ellipses

tejohnso:


ahmedalsudani:
:+1:

PR #356, opened 2016-05-13
Use a separate module for cli options

tejohnso:
@ahmedalsudani @fjvallarino please review.  This gives one place to check for cli params.


installer/cli-options.js
```diff
@@ -0,0 +1,13 @@
+const yargs = require("yargs"),
+params = {}
+
+function reloadOptions() {
```
ahmedalsudani: @tejohnso why not just parse options when the module is loaded?

tejohnso: @ahmedalsudani for testability.

installer/config.js
```diff
@@ -3,17 +3,22 @@ proxy = require("rise-common-electron").proxy,
 network = require("rise-common-electron").network,
 networkCheck = requireRoot("installer/network-check.js"),
 machineId = requireRoot("installer/machine-id.js"),
-electronAppPath = "",
 path = require("path"),
 proxyConfig,
 uiWindow;
 
+function sendUIMessage(msg, data) {
```
ahmedalsudani: @tejohnso why don't we have uiWindow.send do this check? Then we wouldn't need a function in every file to check for a valid window

tejohnso: @ahmedalsudani I wouldn't want to modify electron's webContents object.  
Instead of setting the uiWindow object to webcontents, we could create an object that has a better built-in send that delegates to webContents.send.   But that's for another pr.

ahmedalsudani: Oh, my mistake. I forgot where we get uiWindow from. I meant add a method in ui/controller that checks validity and sends. Let's save it for a future PR though

ahmedalsudani:
:+1:

PR #357, opened 2016-05-13
Use $window.close to close countdown

tejohnso:
This prevents an object has been destroyed uncaught error.

@ahmedalsudani @fjvallarino please review.


ahmedalsudani:
:+1:

PR #358, opened 2016-05-13
Don't set displayid=undefined

ahmedalsudani:
@fjvallarino @tejohnso fix for displayid being set to undefined on test runs.


tejohnso:
:+1:

PR #360, opened 2016-05-13
Set dimensions

tejohnso:


installer/ui/controller.js
```diff
@@ -20,7 +20,7 @@ module.exports = {
   init(displaySettings, stopKey) {
     mainWindow = new BrowserWindow(
     {
-      "width": 600,
+      "width": 800,
       "height": 700,
```
ahmedalsudani: @tejohnso wasn't it supposed to be 800x800?

tejohnso: No I confirmed with Adrian

ahmedalsudani:
:+1:

PR #363, opened 2016-05-13
Don't display buttons while network check is going on

ahmedalsudani:
@fjvallarino @tejohnso fixing something that broke in my last PR


tejohnso:
👍

fjvallarino:
:+1:

PR #364, opened 2016-05-14
Remove no longer used config file entry

fjvallarino:


ahmedalsudani:
@fjvallarino I think we should keep this entry just in case we need to switch back to Chrome

tejohnso:
@ahmedalsudani We can always create the browser entry from the proxy config if we need to.  In fact, we never really needed a separate entry for the browser because it can be mapped directly from the proxy config field.

ahmedalsudani:
:+1:

fjvallarino:
:+1:

PR #365, opened 2016-05-15
Remove multiple proxy related events

tejohnso:
Remove:

 - proxy-fields-update
 - proxy-options-screen-active
 - show-proxy-options
 - checking-network

The proxy service is loaded as a dependency of the main service.  This
way ipc events aren't missed if the proxy screen hasn't loaded yet as is
the case when only the controller has the service as a dependency.

Integration tests are included via electron-mocha.  The main ui
controller file is `require`d and its init function is called to load
the installer UI.  From there ipc events can be passed around as in the
production code, but with none of the production backend processes
running.


installer/config.js
```diff
@@ -109,14 +95,5 @@ module.exports = {
   parsePropertyList,
   setUIWindow(ipcWindow) {
     uiWindow = ipcWindow;
-  },
-
-  setIPC(ipc) {
-    ipc.once("proxy-options-screen-active", ()=>{
-      sendUIMessage("proxy-fields-update", proxyConfig);
```
ahmedalsudani: @tejohnso wouldn't this mean the proxy screen won't have the proxy information that's already saved?

installer/network-check.js
```diff
@@ -71,6 +71,6 @@ module.exports = {
   checkSites() {
     if (cli.params.unattended) {return Promise.resolve();}
     return module.exports.checkSitesWithElectron()
-    .then(module.exports.checkSitesWithJava);
+    .then(module.exports.checkSitesWithJava.bind(null, 1, 15000));
```
ahmedalsudani: We should use global.secondMillis

installer/config.js
tejohnso: @ahmedalsudani It will receive it on the first validation call at start up.

fjvallarino:
:+1:

ahmedalsudani:
:+1:

PR #366, opened 2016-05-16
Network check message has consistent width

ahmedalsudani:
@fjvallarino @tejohnso the message was slowly creeping to the left because it's center-aligned, and its width is changing. In this PR, we append spaces to the end to preserve the width


tejohnso:
@ahmedalsudani where does the magic number 6 come from?

ahmedalsudani:
@tejohnso it's arbitrary. I kept going up from 3 until the pacing looked natural.

tejohnso:
@ahmedalsudani I added a commit.  Ready when you are :+1:

ahmedalsudani:
@tejohnso oops, thanks

PR #371, opened 2016-05-16
Set autostart based on user input

ahmedalsudani:
@fjvallarino @tejohnso I noticed this was never implemented


tejohnso:
:+1:

fjvallarino:
:+1:

PR #373, opened 2016-05-16
Don't lie about setting autostart

tejohnso:
@ahmedalsudani @fjvallarino this was confusing me as to why autostart was "being set" but didn't appear to be working.  The log for setting autostart is the same in autostart.js


ahmedalsudani:
:+1:

fjvallarino:
:+1:

PR #376, opened 2016-05-16
Resize icons from originals instead of setting bakground color manually

fjvallarino:
Resize icons from originals instead of setting bakground color manually (artifacts were showing up on borders)

@tejohnso @ahmedalsudani please review (sorry, this should be the last time)


tejohnso:
:+1:

PR #377, opened 2016-05-17
Fix text overflow and links opening in Installer

ahmedalsudani:
@fjvallarino @tejohnso 

I realized the dimensions were only too small if there were too many error on the display registration page. Instead of changing the resolution, I'm hiding the errors when it's possible


tejohnso:
:+1:

fjvallarino:
:+1:

PR #383, opened 2016-05-17
Use the installer's proxy config within viewer

tejohnso:
Otherwise the system proxy configuration is used, which hasn't been
validated by the installer.

@ahmedalsudani @fjvallarino please review.
We need to force the viewer to use the proxy config that the installer is aware of.  Otherwise there is inconsistency between what is being validated through the network checks and what is being used in viewer.  

At some point we should consider getting rid of manual proxy setup within the installer and see if we can always use the system's proxy settings.  Card created.


fjvallarino:
:+1:

viewer/controller.js
```diff
@@ -2,11 +2,18 @@
...
 const network = require("rise-common-electron").network;
+const proxy = require("rise-common-electron").proxy;
 const config = requireRoot("installer/config.js");
 const version = requireRoot("version.json");
 let BrowserWindow;
 let app;
 let globalShortcut;
+let proxyConfig="http=:";
+
+proxy.observe((config)=>{
```
ahmedalsudani: @tejohnso I'm thinking we should put this observer in `registerEvents`, so we're guaranteed `proxyConfig` will be set when we use it.

tejohnso: @ahmedalsudani I think the opposite.  We want to register the observer at initialization time, then we're sure to catch any proxy changes.

ahmedalsudani: @tejohnso doesn't the observe method call the observer right away?

tejohnso: @ahmedalsudani It does not.  We did that in the ui though.  So we can always do it that way too.  But at that point you're just using it as a getter :)  So I'd probably just use getCurrentConfig() or something.   Anyway....I think it's fine as it is here listening to the events.

ahmedalsudani: @tejohnso I see. I'm still concerned about the race condition we might have here, but let's see what happens

ahmedalsudani:
:+1:

PR #390, opened 2016-05-26
Create graceful shutdown flag file on install

fjvallarino:


tejohnso:
:+1:

tejohnso:
@fjvallarino I don't think we should have merged this yet as it wasn't part of the R13 testing handover.  Also wondering if `echo.` is a typo in the batch file.

PR #403, opened 2016-06-06
Escape '#' in installer UI URL

ahmedalsudani:
@tejohnso ready for review. It doesn't look like any other characters would be problematic.

> Logon names can't contain certain characters. Invalid characters are " / \ [ ] : ; | = , + * ? < >

> Logon names can contain all other special characters, including spaces, periods, dashes, and underscores. But it's generally not a good idea to use spaces in account names.

From https://msdn.microsoft.com/en-us/library/bb726984.aspx


tejohnso:
:+1:

PR #406, opened 2016-06-06
Update proxy wording and highlight Verify/Save

tejohnso:
@ahmedalsudani please review
proxy


ahmedalsudani:
@tejohnso I like the primary button change. I'm thinking we should move verify and safe to the left to go along with the convention in the rest of the Installer (the default "next step" button is always to the right)

tejohnso:
@ahmedalsudani you mean move it to the right then?

tejohnso:
@ahmedalsudani updated
proxy

ahmedalsudani:
:+1: once R14 is merged

tejohnso:
Deleting to PR onto master instead of R14

PR #411, opened 2016-06-09
Disable pinch to zoom unless specifically required by user

fjvallarino:
requireRoot is redefined because it's not available in this context. I tried avoiding this redefinition, but config.js uses that function. Since the flag for enabling pinch to zoom is read from the .ini file, the alternative was re creating a function for reading the config file.

The branch was created as a "release" because I needed to test on Windows too, and the builds I generate on my local Linux box fail on shortcut creation.


viewer/ipc.js
```diff
@@ -1 +1,12 @@
...
+};
+
+const config = requireRoot("installer/config.js");
+const displaySettings = config.getDisplaySettingsSync();
+const webFrame = require("electron").webFrame;
+
+if(displaySettings.enablePinchToZoom !== "true") {
+  // Disable pinch-to-zoom
+  webFrame.setZoomLevelLimits(1, 1);
+}
```
ahmedalsudani: @fjvallarino how about we add `global.requireRoot = null` at the end?

ahmedalsudani:
:+1:

tejohnso:
:+1:

PR #412, opened 2016-06-09
Add script for config file manipulation

tejohnso:
@ahmedalsudani @fjvallarino I'm still testing this but I wanted to get some notes down right now.  

 - The current production *remote-components-win.cfg* has ^M line endings, which aren't going to be carried forward.  So the diff is going to be a change on every line for that file.

 - The diff is using -rcs for RCS style formatting.  It seems to be more friendly in *sh* which node is using for spawn.


tejohnso:
@ahmedalsudani @fjvallarino ready for review.  Use `node change-config-files.js` to see help screen.

fjvallarino:
@tejohnso looks good :+1:

ahmedalsudani:
:+1:

PR #413, opened 2016-06-09
Allow users to continue even if invalid OS/Arch combination

ahmedalsudani:
@fjvallarino @tejohnso code is ready for review. Integration tests are being added.


ahmedalsudani:
@tejohnso @fjvallarino finished adding tests. This is the final version if it passes reviews.

installer/ui/js/services/main-process.js
```diff
@@ -42,6 +42,13 @@ angular.module('riseInstaller')
       });
     });
 
+    ipc.on("set-recoverable-error", (evt, message)=>{
+      SharedState.set({
```
tejohnso: @ahmedalsudani instead of using sharedstate can you create a recoverable-error-screen-service and have main-process use that service?  Similar to the way ProxyService is used.   I think we should move away from the SharedState overuse.

ahmedalsudani: @tejohnso Hmm. I agree that's a better solution, but it's still iffy. The correct way to do it would be to wrap the service declarations in IIFEs and put the IPC handler outside the service. That way, you'll only register the handler once, and its code block will be immediately evaluated (even if it's not injected into main-process)

tejohnso: @ahmedalsudani I'm not sure why you'd need to wrap the services.  As it is, proxyservice is only registering the handler once. The constructor portion of the service doesn't get re-executed on each injection.    The ipc handler for a service should be in the service so that everything to do with that service is contained within it.

test/integration/ui/recoverable-error-it.js
```diff
@@ -0,0 +1,26 @@
...
+    });
+
+    it("switches to Intro tab if error message is null", ()=>{
+      testUtils.send("set-recoverable-error", "error text");
+      testUtils.send("set-recoverable-error", null);
+      return testUtils.waitForElement('span[ng-controller*="IntroCtrl"]');
+    });
+
+  });
+};
```
tejohnso: @ahmedalsudani We should have a test that confirms "attended-ready" is sent to ipc when the *Continue* button is clicked.

ahmedalsudani: @tejohnso good catch. Added

PR #415, opened 2016-06-10
Use test company id for integration test

tejohnso:
@fjvallarino please review - we're creating a bunch of test display id's under the rise vision company so I created a test company instead.


fjvallarino:
:+1:

PR #419, opened 2016-06-14
Add electron-prebuilt to CircleCI build process

fjvallarino:


tejohnso:
@fjvallarino this is going to install a version of Electron that might not match the version we use for the build.  Maybe we need an *electron-version* file so that they don't get out of sync.

Or maybe could add electron-prebuilt back into package.json and have build.sh get the version from there.

fjvallarino:
@tejohnso I agree, but this is what we are suggesting developers do on their local environment, so I think  we should also change the documentation.

tejohnso:
@fjvallarino Sounds good.

ahmedalsudani:
@fjvallarino @tejohnso what was the problem with having `electron-prebuilt` in our dev dependencies?

fjvallarino:
@tejohnso @ahmedalsudani updated to restore electron-prebuilt on package.json, use that same value from build.sh and remove README references to manually installing electron-prebuilt.

tejohnso:
:+1: 
@ahmedalsudani It's back in dev dependencies, but instead of specifying latest version, it's now pinned to a specific electron version and the build is tied to the same version so we're building, testing, and developing all on the same version.

PR #420, opened 2016-06-14
Support user provided resolution

fjvallarino:


viewer/controller.js
```diff
@@ -117,7 +124,11 @@ module.exports = {
             "nodeIntegration": false,
             "webSecurity": false
           }
-        });
+        }, customResolutionSettings));
+
+        if(customResolution) {
+          viewerWindow.setSize(Number(displaySettings.screenwidth), Number(displaySettings.screenheight));
+        }
```
tejohnso: @fjvallarino Why can't we set the x: y: custom resolution when the browser window is created instead of setting them to zero and then calling setSize?

fjvallarino: @tejohnso because it was not working. The size would get reduced to whatever the main display's size is. Manually calling setSize after window creation works fine.

tejohnso:
:+1:

ahmedalsudani:
:+1:

PR #421, opened 2016-06-15
Chore/component tool update

tejohnso:
@ahmedalsudani please review reminders as per recent discussions around v2 releases.


ahmedalsudani:
@tejohnso looks good, but I'm wondering: why not use `/tmp`?

PR #423, opened 2016-06-16
Use full os description in BQ and displays app

tejohnso:


main.js
```diff
@@ -14,7 +14,7 @@ version = requireRoot("version.json"),
 displaySettings;
 
 var externalLogger = require("rise-common-electron").externalLogger
-  (network, platform.getOS(), platform.getArch(), version);
+  (network, platform.getOSDescription(), platform.getArch(), version, platform.getOSDescription());
```
ahmedalsudani: @tejohnso why would we have two parameters if we're going to pass the same argument?

tejohnso:
@ahmedalsudani updated

ahmedalsudani:
:+1:

fjvallarino:
:+1:

PR #424, opened 2016-06-16
Use rise-common-electron v1.0.0

tejohnso:


ahmedalsudani:
@tejohnso why not use commit hashes? Using tags will create extra overhead

tejohnso:
@ahmedalsudani Agreed, not worth the overhead.  @fjvallarino ?

fjvallarino:
@tejohnso @ahmedalsudani makes sense

PR #430, opened 2016-06-20
Retry fetching config json

ahmedalsudani:
@tejohnso ready for review


tejohnso:
@ahmedalsudani We should add a unit tests for the failure cases as well.

ahmedalsudani:
@tejohnso those are already covered. I added a callCount check for each failure test

tejohnso:
:+1:

PR #433, opened 2016-06-21
Don't force always on top

tejohnso:
Some users like to obscure their presentations with other media


ahmedalsudani:
:+1:

fjvallarino:
:+1: 

@tejohnso great description!

PR #434, opened 2016-06-22
Import installer specific code from common

fjvallarino:


tejohnso:
:+1:

installer/config.js
```diff
@@ -91,8 +97,71 @@ module.exports = {
...
+  },
+  updateDisplaySettings(newConfiguration) {
+    if (typeof newConfiguration != "object") {
+      throw new Error("Incorrect configuration type");
+    }
+    const configFile = module.exports.getDisplaySettingsPath();
+    const currentConfiguration = module.exports.parsePropertyList(platform.readTextFileSync(configFile));
+    const updatedConfiguration = Object.assign(currentConfiguration, newConfiguration);
+    const updatedConfigurationText = Object.keys(updatedConfiguration).reduce((acc, curr)=>{
+      if (updatedConfiguration[curr] !== null) {
```
ahmedalsudani: @fjvallarino one thing to be aware of: this block will be entered if `updatedConfiguration[curr] = undefined` and we'll end up with ${curr}=undefined\n`

ahmedalsudani:
:+1: 

don't forget to lock down `rise-common-electron` to a commit hash :smile:

installer/config.js
fjvallarino: @ahmedalsudani this is code imported, unmodified, from rise-common-electron (https://github.com/Rise-Vision/rise-common-electron/pull/54). Since we're reading the keys of an object, that scenario should not actually happen.

ahmedalsudani: @fjvallarino the condition uses `!=` (not `!==`) in the original version, which returns true if `updatedConfiguration[curr] ` is undefined.

I agree the scenario should not happen, but it's not inconceivable.

fjvallarino:
@tejohnso @ahmedalsudani restored != null check and updated jshintrc

ahmedalsudani:
:+1:

PR #436, opened 2016-06-29
Fix/use browser online check

tejohnso:
@ahmedalsudani @fjvallarino I decided to delegate the entire online check headache to the browser's built-in `navigator.onLine` functionality.  

Removed the special browserwindow for the no-network screen.  Instead we load the viewer's browser window, and change what url we load into it depending on network connectivity.


fjvallarino:
@tejohnso :+1: 

Looks good. I remember having some issues with navigator.onLine on the offline player. We tried watching the property for changes and it did not really work (or it took way too long for the value to change). Probably not an issue anymore.

PR #438, opened 2016-07-06
Update amazon network check endpoint

tejohnso:
@ahmedalsudani @fjvallarino This will reduce problems with network checks on proxied machines.


fjvallarino:
:+1:

ahmedalsudani:
@tejohnso I realize this will work, but a more general solution is to accept a wider range of status codes. I'm looking into whether that's possible.

tejohnso:
@ahmedalsudani I'm going to merge this for the time being - it's not going to cause a regression even if we end up implementing a more general solution.

PR #439, opened 2016-07-07
Only fail if we receive a 403 or can't connect

ahmedalsudani:
@fjvallarino @tejohnso ready for review.

I am not sure what the best way to proceed with proxy is, but we certainly shouldn't assume there's a network issue if we receive a 5xx status. This PR changes the installer behavior so proxy checks only fail if we receive a 403 or if the network is down.


tejohnso:
@ahmedalsudani How are you so sure that 403 is the only error we need to worry about?

ahmedalsudani:
@tejohnso I am not, but we certainly should not be interpreting the majority of error codes as an indication we need a proxy.

I'm still of the opinion that setting up a proxy should be a user-initiated action.

tejohnso:
@ahmedalsudani We're not taking it as an indication that we need a proxy, but we're presenting the option (which can be skipped) just in case proxy is the problem.  

I don't think we should be interpreting the majority of error codes as an indication that the network is fine.  We need to let the user know that there's something going on.  At very least we'd want to fail on HTTP Error 407 Proxy authentication required.  But it's safer to present the user with the option on any error and give them an opportunity to think about it or contact us.

tejohnso:
:+1:

PR #444, opened 2016-07-23
Add screenshot capture

fjvallarino:

A few comments:

- We are not sending anything when the capture/upload could not be performed
- I guess the screenshot capture should be retried
- We should think about an easier way to test messaging handlers. In this case, I mocked registerHandler to get an instance of what I wanted to test, but the condition function was left untested. Maybe adding some easily mockable method for simulating messages.
- This branch was just rebased from master; I'm not sure why the already existing commits were kept


tejohnso:
@fjvallarino 

 - Why don't we have an integration test for `uploadBuffer` to confirm it gets into GCS as expected.  

 - Why not send a `screenshot-failed` message back from the handler if there was a problem?

fjvallarino:
@tejohnso 

- The integration test was not added since I understood we were going to validate the complete process with an e2e test.
- I did not find a message with that name on the messaging server. Does it exist, or will the server relay any message a client sends it?

tejohnso:
@fjvallarino 

 - Okay, we should cover that case with the e2e tests
 - There's no such message server-side at this point, but one should be added.  In the meantime you can send it.

ahmedalsudani:
@fjvallarino good to merge once it's rebased and Tyler's comment is addressed.

fjvallarino:

ahmedalsudani:
:+1:

tejohnso:
:+1:

PR #445, opened 2016-07-28
Harden Continuous Delivery

ahmedalsudani:
@tejohnso @fjvallarino ready for review

- check that remote files have been modified
- check that remote files match local files
- use signed executable for E2E tests on Windows


tejohnso:
@ahmedalsudani this seems okay, but we're still fully deployed by the time we check the remote files right?  So we could still end up in the same scenario where we have the wrong acl on a production file for example.

ahmedalsudani:
@fjvallarino @tejohnso updated. Ready for review once again

tejohnso:
:+1:

PR #448, opened 2016-08-08
Report Electron-resolved proxy to BQ on startup

ahmedalsudani:


fjvallarino:
:+1:

installer/telemetry.js
```diff
@@ -0,0 +1,20 @@
...
+    mainWindow = _mainWindow;
+    module.exports.runProxyExperiment();
+  },
+  reportProxy(url, proxy) {
+    log.external("resolved proxy result", `${url}: ${proxy}`);
+  },
+  runProxyExperiment() {
+    const session = mainWindow.webContents.session;
+    const urls = ["risevision.com", "127.0.0.1", "192.168.1.1", "10.0.0.1"];
+    urls.map((url)=>{
```
tejohnso: @ahmedalsudani This maps the url array to a bunch of undefineds and then does nothing with the mapping.  I think forEach would be more appropriate.

tejohnso:
@ahmedalsudani One comment to consider - looks good :+1:

installer/telemetry.js
ahmedalsudani: @tejohnso ah, good point. Going to replace with a forEach before merging.

PR #449, opened 2016-08-15
Save Player config to BQ

fjvallarino:


tejohnso:
:+1:

viewer/controller.js
```diff
@@ -49,6 +58,14 @@ function registerEvents(window) {
     }
   });
 
+  ipc.on("viewer-message", (evt, message)=>{
+    configLogger.logClientInfo(message.viewerVersion)
+      .catch((err)=>{
```
ahmedalsudani: @fjvallarino I think we need to retry on failure. We have had instances where logging to BQ fails because the display hasn't yet connected to the network.

fjvallarino:
@tejohnso @ahmedalsudani updated. Now it retries logging client config every 30'', with a 10'' timeout, a total of 10 times. This amounts to retrying for about five minutes as requested.

fjvallarino:

viewer/controller.js
```diff
@@ -49,6 +58,14 @@ function registerEvents(window) {
     }
   });
 
+  ipc.on("viewer-message", (evt, message)=>{
+    platform.runFunction(configLogger.logClientInfo.bind(null, message.viewerVersion), 10, 10000, 30000)
```
tejohnso: @fjvallarino I think we should check to see what the message is before assuming we need to logClientInfo.  I guess right now there's only the one message but even so it would be more clear.

tejohnso: We can use this for the watchdog message as well rather than having a separate message posting function specifically for the watchdog. @ahmedalsudani

PR #455, opened 2016-08-25
Feature/duplicate

tejohnso:
@ahmedalsudani @fjvallarino please review.  Testing on linux and windows but I'm dealing with a problem on windows right now.  It seems to be fixed by avoiding the use of ASAR but that's not a great solution.


fjvallarino:
:+1:

installer/screenshot.js
```diff
@@ -33,5 +46,9 @@ module.exports = {
           }
       });
     });
-  }
+  },
+  setWindow(win) {
+    viewerWindow = win;
+  },
+  setDisplayId(id) {displayId = id;}
```
ahmedalsudani: @tejohnso minor: we can get displayId without having it injected (as we do [here](https://github.com/Rise-Vision/rise-launcher-electron/pull/455/files#diff-879671d6a9ad41e202f04eb2cdda2c99R14))

ahmedalsudani:
Addressing the comment above is up to you, @tejohnso.

Once tests are passing, :+1:

PR #459, opened 2016-08-31
Player fetches content and sends it to Viewer

ahmedalsudani:
@tejohnso @fjvallarino ready for review

This undoes some of the work previously done. Instead of Viewer directly listening for messages, we now listen on the main process side. When the content update message is received, Player fetches the configuration and sends it to Viewer. What is not yet being done:
  - using the saved configuration
  - sending the saved configuration to Viewer on startup

Room for improvement: retry if we fail to fetch content.


tejohnso:
@fjvallarino ready for review

viewer/content-loader.js
```diff
@@ -0,0 +1,34 @@
...
+        msg: "content-update",
+        newContent: content
+      });
+    });
+  }
+};
+
+function getContent() {
+  return gcs.getFileContents(contentPath)
+  .catch((err)=>{
```
fjvallarino: @tejohnso maybe we could reject here in order to avoid passing invalid content to Viewer (unless Viewer ignores the "no-content" flag). Does it make sense to retry a few times when fetching new content?

tejohnso: @fjvallarino Let's throw and log an external error for now rather than trying to have viewer deal with it.  Go ahead and make the change.

`gcs.getFileContents()` will handle retries.

fjvallarino:
:+1:

tejohnso:
👍

PR #461, opened 2016-09-05
Return created window from Viewer controller

fjvallarino:
Previously index.launch was equal to controller.launch:
`index.launch: controller.launch`

This meant controller.launch's response (the created viewerWindow) was returned and subsequently used by other parts of Player (e.g., screenshot requests). This PR restores that functionality.

@tejohnso please review


tejohnso:
:+1:

PR #462, opened 2016-09-06
Fix GCS integration test timeout

fjvallarino:
The mocked methods were not resolving/rejecting to the expected values, which caused the tests to never complete their run. Why those tests work locally is completely beyond my understanding.

@alexey-rise please review

@tejohnso FYI


test/integration/gcs.js
```diff
@@ -86,7 +86,7 @@ describe("GCS", function() {
     return getCurrentGeneration()
     .then((currentGeneration)=>{
       let fakeCommand = Math.random();
-      simple.mock(network, "httpFetch");
+      simple.mock(network, "httpFetch").resolveWith({ status: 304 });
```
tejohnso: @fjvallarino we don't want to mock out the fetch here.  This test should hit the actual gcs endpoint and it should return 304 if the generation is the same as was determined from getCurrentGeneration()

fjvallarino: @tejohnso makes sense. Something is causing this to fail on Circle, but not locally. I'll keep reasearching.

PR #464, opened 2016-09-07
Fail when request for json config files is 403

tejohnso:
The script is meant to confirm that the config files are publicly
downloadable.  Without `-f` curl exits 0 even on a 403.

@fjvallarino please review.  The latest master build was successful even though the linux 64 json file was not publicly downloadable.


fjvallarino:
As long as we're ok with this:

`This method is not fail-safe and there are occasions where non-successful response codes will slip through, especially when authentication is involved (response codes 401 and 407).`

:+1:

PR #470, opened 2016-09-14
Add failure handling on startup content fetch

tejohnso:
@fjvallarino please review

GCS module returns local data on multiple 404 or fetch failure (eg offline)
Viewer logs externally if there's no data.


fjvallarino:
:+1:

PR #471, opened 2016-09-19
Fix/circular json

tejohnso:
@alexey-rise please review

Watchdog starts earlier in the startup process, and timeout is 60s.

@rodrigopavezi over to you as well as discussed in SoS


rodrigopavezi:
@tejohnso looks good 👍  cheers

PR #473, opened 2016-09-23
Improve duplicate display id handling

tejohnso:
Don't kill the player if the duplicate connection is from the same
machine.
@fjvallarino please review


fjvallarino:
:+1:

PR #478, opened 2016-09-29
Don't show no-network on fail-load

tejohnso:
did-fail-load fires for all external requests within the page and if any
fail the no-network screen will be displayed.
See Issue#477

@fjvallarino please review


fjvallarino:
:+1:

PR #487, opened 2016-10-06
Use master branch of node-electron-proxy-agent

fjvallarino:
@tejohnso please review

@rodrigopavezi FYI. In case you push changes to your branch and it fails to build, make sure you rebase to latest staging/alan-test (after this PR is merged, of course).


fjvallarino:
This will be handled in a separate PR which updates versions of other dependencies.

PR #492, opened 2016-10-10
Reset log files on startup

tejohnso:
@fjvallarino please review


fjvallarino:
@tejohnso I think we should only reset them if they grow beyond certain size. That way we will be able to see what happened to the user's machine when some issue arises.

tejohnso:
@fjvallarino updated

fjvallarino:
:+1:

PR #497, opened 2016-10-14
Remove unnecessary logging

fjvallarino:
@tejohnso please review


installer/network-check.js
```diff
@@ -79,7 +79,7 @@ module.exports = {
 
     return Promise.all(siteConnections)
       .then((res)=>{
-        log.file(JSON.stringify(res));
+        log.file("Electron network checks passed: " + JSON.stringify(res));
```
tejohnso: @fjvallarino isn't this now going to log
>Electron network checks passed: [null, null, ...]

Making it look like nothing passed?

fjvallarino: @tejohnso yeah; I only added the "Electron network checks passed: " to know where that log line came from. We don't actually need the result of the computation. In fact, we don't really need the log.

fjvallarino: @tejohnso I meant "passed" as in "the network checks were successful"

tejohnso: @fjvallarino I think we should remove "passed" and change it to *"Electron network checks:"*
Other than that, :+1:

PR #498, opened 2016-10-19
Force resolve even if viewer load times out

tejohnso:
@fjvallarino This way we still log the issue but we don't end up showing the no-network screen.


fjvallarino:
:+1:

PR #500, opened 2016-11-03
Don't assume stringifiable error

tejohnso:
@fjvallarino please review.  We're getting `"Unexpected token < in JSON at position 0"` as our error detail in some instances.
@alanclayton FYI


fjvallarino:
Should we use ```require("util").inspect``` instead? If you feel it's not worth it, it's ok for me.

:+1:

tejohnso:
@fjvallarino updated

fjvallarino:
Thanks! :+1:

PR #502, opened 2016-11-08
Use rise-common to v1.0.1

fjvallarino:
This updated version fixes the proxy issue and increases the download timeout

@Rise-Vision/apps please review


tejohnso:
@fjvallarino Did the versioning change correct the conflict issue with RC?
:+1:

PR #509, opened 2017-01-10
Fix screenshot upload

fjvallarino:
@tejohnso please review

@alex-deaconu @ezequielc FYI


tejohnso:
@fjvallarino why did you have to bring in another http library just for this? Do you know what was causing the problem?  Should we move everything off of node-fetch and use request?  It'd be better if we weren't using two different libraries.

fjvallarino:
@tejohnso node-fetch is not working correctly for uploading files to GCS (it seemed to work in the past, but it clearly does not work now). It successfully creates the file, but does not upload the content (you get a zero length file). I tested the signed URL provided by Core with curl and it succeeded, which prompted me to try a different library. Since we already had request in our dependencies (at some point we evaluated moving from node-fetch to request), I just used that one, and it worked without issues (even behind a proxy). The request library is also used by rise-cache-v2.

We currently use node-fetch for regular requests, not for uploads. For downloading upgrades (a big file), we have custom download code using node's http module.

I agree that using a single library would be best, but some code in Player relies on node-fetch's API. I remember trying to simulate node-fetch's API using request in the past by providing an object with { text(), json, status }, but I'm not sure it covers all the cases.

tejohnso:
@fjvallarino do you think we can use *request* and drop *node-fetch*?  I'm not sure it's worth the work but I think this situation qualifies as technical debt.

fjvallarino:
@tejohnso I'm sure we can, although I'm in the same situation regarding if it's worth doing it now. I'm worried we may break something unexpected.

PR #513, opened 2017-02-10
Use latest rise common with offline external logging

tejohnso:
@fjvallarino @stulees @rodrigopavezi for review.


stulees:
@tejohnso 👍

PR #515, opened 2017-02-14
Update version of Rise Cache

stulees:
- Fixes uncaughtException regarding `fs.unlink()` function


stulees:

PR #519, opened 2017-03-01
Use local viewer

tejohnso:
@fjvallarino @rodrigopavezi please review.  We just copy viewer from GCS and it's built as part of the player bundle.


fjvallarino:
:+1:

PR #524, opened 2017-03-28
Offline enabled with various Widgets

stulees:
Bundles the following widgets:
- Image
- Video
- Text
- Spreadsheet

Only uses offline/bundled widgets when offline Viewer is used


stulees:

PR #527, opened 2017-04-05
Enable keyboard navigation during installation

tejohnso:
Allows tabbing for navigation through buttons and adds highlight to indicate focused button


stulees:
@tejohnso 👍

PR #540, opened 2017-04-24
Upgrade Electron version

stulees:


stulees:

package.json
```diff
@@ -17,7 +17,7 @@
     "camelcase": "^2.1.0",
     "electron-mocha": "^2.0.0",
     "electron-packager": "8.6.0",
-    "electron-prebuilt": "1.3.5",
+    "electron-prebuilt": "1.4.13",
```
tejohnso: Why not `"1.6.6"` - latest version

stulees: @tejohnso `"1.3.5"` was previously the npm version so I followed suit. If we want `"1.6.6"` then we need to point to Github URL, otherwise going on version number only it needs to be latest `npm` release which is [1.4.13](https://www.npmjs.com/package/electron-prebuilt).

tejohnso: Looks like this needs to be renamed to `electron`

[... the electron-prebuilt name is deprecated, and will only be published until the end of 2016.](https://github.com/electron-userland/electron-prebuilt)

stulees:
@tejohnso Package name/version revised in last commit, thanks

tejohnso:
Added a commit for `build.sh`

PR #543, opened 2017-04-29
One-time relaunch of viewer when detected offline to online

stulees:
- Added viewer `window-bindings` module to maintain viewer window reference for sending content to viewer
- Revised launcher and viewer controller to use window bindings for setting viewer window reference and revised viewer content-loader to use window bindings to send content to viewer
- Viewer controller handles "online-status-change" to check if status is going from offline to online and if it hasn't reloaded before
- Added `reload()` function to viewer controller for destroying current viewer window, launch and load online viewer and send content
- Unit tests added and revised


viewer/controller.js
```diff
@@ -60,6 +67,19 @@ function registerEvents(window) {
       dataHandlerRegistered = true;
     }
   });
+
+  ipc.on( "online-status-changed", ( evt, status ) => {
+    if ( !initialLoadOnline && status === "online" && !reload ) {
```
tejohnso: I think this check could be simpler without introducing two variables:

`if (status === "online" && window.webContents.getURL().startsWith("file"))`

stulees: @tejohnso If we remove the conditional use of those flags, it can allow for unnecessary reloading more than once if their display network goes offline -> online -> offline -> online .... etc. The idea was just to reload one time when starts up offline and goes to online. Thoughts?

tejohnso: offline->online | viewer is reloaded to the online viewer
online->offline | nothing changes because status !== "online"
offline->online | nothing changes because we're already using online viewer (no *file* url)

Is there a case missing?

stulees: @tejohnso I overlooked that last case (no *file* url), you're correct. Makes sense to make the change, thanks.

stulees:
@tejohnso Please review last commit, thanks.

stulees:
@tejohnso Please review last commit, thanks

tejohnso:
:+1:

PR #552, opened 2017-06-01
Disable offline play product

stulees:


stulees:
@tejohnso Ready for review

installer/launcher.js
```diff
@@ -41,20 +40,15 @@ module.exports = {
       return watchdog.init();
     })
     .then(()=>{
-      const offlineSubscribedPromise = offlineSubscriptionCheck.isSubscribed();
-
-      if (!onlineDetection.isOnline() ) {
-        return offlineSubscribedPromise;
-      }
-
-      return true;
```
tejohnso: Instead of changing this whole section, it might be better to just add a `return false;` to  offlineSubscriptionCheck.isSubscribed() and leave all of this as is.  Otherwise looks good.

stulees:
@tejohnso Agreed, better to just return false in subscription check. Made the changes and tested, works good.

stulees:
@tejohnso It turned out jshint gave lots of trouble when just returning false in `isSubscribed()` and I also had to turn off tests for offline-subscription-check.js. I've put the code back to the way it was for previous review as I think it's better.

PR #553, opened 2017-06-01
Feature/expected ready count

tejohnso:
@stulees I set the test up so that "abcdWIDGETSTRINGefghWIDGETSTRING" would result in two expected ready counts since I'm pretty sure each widget loaded should send a separate ready event.  Even if it's the same type of widget.


stulees:
@tejohnso 👍 Looks good

PR #561, opened 2017-06-19
Fix ui-integration tests

stulees:


stulees:
@rodrigopavezi Last couple builds to master failed, I believe this should fix it.

PR #564, opened 2017-06-22
Identify latest player version

tejohnso:
http://install-versions.risevision.com/latest-version


tejohnso:
@stulees I added a commit that should work. We won't know for sure until we try to merge it.

PR #569, opened 2017-06-29
Register custom protocols, handle protocol requests, grant protocol privileges

stulees:


installer/main-controller.js
```diff
@@ -51,11 +52,34 @@ function scheduleDiskSpaceCheck() {
   }, 4 * 60 * global.secondMillis);
 }
 
+function schemeHandler(request, callback, secure) {
```
tejohnso: Why do we need a `secure` boolean here?  The `request` will contain the url which contains the scheme which will end in `s` if it's a secure request.

installer/main-controller.js
```diff
@@ -51,11 +52,34 @@ function scheduleDiskSpaceCheck() {
...
+    postData;
+
+  if (request.uploadData) {
+    postData = request.uploadData[0].bytes.toString();
+  }
+
+  if (request.method === "POST" || request.method === "PUT") {
+    redirect.uploadData = {
+      contentType: "application/json",
+      data: postData
```
tejohnso: This could be 

```
data: request.uploadData ? request.uploadData[0].bytes.toString() : ""
```

rather than having a separate if statement up on l62

installer/main-controller.js
```diff
@@ -51,11 +52,34 @@ function scheduleDiskSpaceCheck() {
...
+    };
+  }
+
+  redirect.url = (secure) ? `https://${url}` : `http://${url}`;
```
tejohnso: Couldn't this be defined as part of the object literal on l57?

stulees:
@tejohnso Changes made, please review, thanks

stulees:
@tejohnso I've simplified the scheme handler as I found an issue when testing RC urls regarding `/files` endpoint as the urls for that endpoint include a query param `?url=` and was breaking the scheme handler. Please review last commit, thanks.

PR #574, opened 2017-07-07
Log not online before viewer launch

tejohnso:
We want to know if a player is starting online or offline.  We do not
want to prevent the offline log just because the offline play
subscription is preventing viewer from launching.


stulees:
@tejohnso 👍

PR #576, opened 2017-07-07
Added offline_subscription to player configuration

rodrigopavezi:
@tejohnso please review. cheers


installer/loggers/config-logger.js
```diff
@@ -33,19 +35,24 @@ module.exports = {
         mac: network.getMAC(),
         serial_number: config.getSerialNumber()
       };
-      let playerConfigStr = module.exports.stringify(playerConfig);
 
-      if(!config.fileExists("display_config.json") || config.readFile("display_config.json") !== playerConfigStr) {
-        playerConfig.ts = nowDate.toISOString();
+      offlineSubscriptionCheck.isSubscribed().then((isSubscribed) => {
```
tejohnso: I think you need a `return` here otherwise the `logClientInfo()` promise is going to be resolved before everything below is completed.

test/unit/loggers/config-logger.js
```diff
@@ -32,9 +34,11 @@ describe("Config logger", ()=>{
 
     return configLogger.logClientInfo(message)
       .then(()=>{
-        assert(bqClient.insert.called);
-        assert(config.writeFile.called);
-        assert.equal(bqClient.insert.lastCall.args[1].viewer_version, "viewerVersion");
+        setTimeout(()=>{
```
tejohnso: This timeout should not be necessary if the `logClientInfo()` promise is returned properly as per previous comment.

test/unit/main-controller.js
```diff
@@ -122,7 +122,7 @@ describe("mainController", ()=>{
       setTimeout(()=>{
         assert(mocks.app.quit.called);
         done();
```
tejohnso: Returning the promise is all that's necessary for the async test in mocha.  There should not be a promise returned and a `done` callback used.

rodrigopavezi:
@tejohnso thanks, changes made. cheers

PR #582, opened 2017-07-12
Remove starting log until after ready

tejohnso:
As in #3735b32.  Necessary with use of net module.

@stulees Did you deliberately add that started log call back into the init() function as part of https://github.com/Rise-Vision/rise-launcher-electron/commit/f22011d498695676376de5e66ae15ed2e17d7e88?  I'm guessing it might have been the result of a merge.  If it was deliberate, we'll have to discuss because it's causing a problem with the `net` module.


stulees:
@tejohnso Looks to be the result of a merge. Don't recall adding that in, doesn't have any significance to the other changes I made.

PR #591, opened 2017-07-28
Consistently return a promise in subscription check

tejohnso:
When offline, we return the file state, but it should be a promise
because it is expected to be a promise in [config-logger](https://github.com/Rise-Vision/rise-launcher-electron/blob/master/installer/loggers/config-logger.js#L39).

- added integration test


rodrigopavezi:
@tejohnso looks good 👍  cheers

PR #607, opened 2017-09-11
Use latest RC

tejohnso:
Use latest RC which does not do any encoding.


stulees:
@tejohnso I see `fdfb376` as latest commit 

image

tejohnso:
@stulees That's the one specified in this change

stulees:
@tejohnso Okay, just double checking as I see an extra 2 digits specified.

PR #608, opened 2017-09-18
Send message to RC to quit when Player is upgrading

stulees:


stulees:
@rodrigopavezi @tejohnso Please review, thanks

PR #617, opened 2017-10-16
Don't init loggers until after ready

tejohnso:
On rare occasions Electron's *ready* event isn't called for over ten
seconds after startup. Loggers set a 10s timer to push their unlogged events after
10s. If they do this immediately after startup, the unlogged event push
might occur before Electron's *ready* event has been called. Then the
*got* module will throw because it is being used to push the unlogged
events and it can only be used after Electron's *ready* event.


stulees:
@tejohnso Makes sense, looks good 👍

PR #618, opened 2017-10-18
Replace # in user name with encoded value %23

stulees:


tejohnso:
Looks good @stulees

tejohnso:
Will be included in modularized player

PR #619, opened 2017-10-19
Implemented the receiving of player messages

rodrigopavezi:
@tejohnso @stulees please review. cheers


installer/lms-client.js
```diff
@@ -0,0 +1,23 @@
+const moduleCommon = require("common-display-module");
+let message = {
```
stulees: @rodrigopavezi This `message` variable doesn't appear to be used for anything

installer/main-controller.js
```diff
@@ -75,7 +99,6 @@ function readyHandler() {
     return platform.killExplorer()
       .then(()=>explorerKilled = true)
       .then(()=>moduleController.startModule("player"))
-      .then(()=>mainWindow && !mainWindow.isDestroyed() && mainWindow.close())
```
stulees: @rodrigopavezi Can you clarify why this is being removed?

rodrigopavezi: @stulees because we need launcher to be running so player can send messages to it. It will only quit when player tells it to do so. cheers

stulees: @rodrigopavezi thanks

installer/main-controller.js
```diff
@@ -52,6 +53,27 @@ function scheduleDiskSpaceCheck() {
   }, 4 * 60 * global.secondMillis);
 }
 
+function handleLMS() {
+  lmsClient.receiveMessages().then((receiver)=>{
```
tejohnso: I thought we removed the promise from the receiver - in which case the *then* isn't going to work and it would just be receiveMessages().on()

But similar to the other PR, I don't think we need to expose the client to the display modules. All they should have to do is moduleCommon.receiveMessages(handler)

installer/lms-client.js
```diff
@@ -0,0 +1,20 @@
...
+    moduleCommon.connect("installer").then((lmsClient)=>{
+      resolve(lmsClient.receiveMessages());
+    });
+  });
+}
+
+module.exports = {
+  broadcastMessage,
+  receiveMessages
+};
```
tejohnso: This can be removed now right?

rodrigopavezi: @tejohnso good catch, sorry missed this one. cheers

PR #623, opened 2017-10-20
Fix player module name at launch

tejohnso:


installer/main-controller.js
```diff
@@ -95,7 +95,7 @@ function readyHandler() {
     log.debug("main-controller: on launch");
     return platform.killExplorer()
       .then(()=>explorerKilled = true)
-      .then(()=>moduleController.startModule("player"))
+      .then(()=>moduleController.startModule("player-electron"))
```
tejohnso: This should probably start all modules rather than specifying player-electron

rodrigopavezi: @tejohnso I think you are right, we should start all of them there. And for sure LMS should be the first. cheers

tejohnso:
@rodrigopavezi see latest change. The startup is async though.

PR #624, opened 2017-10-20
Create extraction dir before module extraction

stulees:


stulees:
@tejohnso Good catch, thanks

PR #627, opened 2017-10-23
Ensuring background modules are started before foreground modules

stulees:


stulees:
@tejohnso I've removed a unit test and had to disable an installer integration test due to no longer restarting a module once extracted. I've removed the module restart upon extraction because we need precedence to background modules always starting before foreground, as per the changes in this PR. Need a re-think on how we upgrade and restart modules with this rule now. Thoughts? @rodrigopavezi FYI

installer/main-controller.js
```diff
@@ -370,7 +371,14 @@ module.exports = {
...
       }
     })
+    .then(()=>{
+      if (options && options.startBackground) {
+        moduleController.startBackgroundModules();
+      }
+    })
+    .then(()=>{})
     .then(()=>{log.all("install complete", "", 100);})
+    .then(()=>{})
```
tejohnso: What are these empty functions for?

installer/installer.js
```diff
@@ -102,7 +101,8 @@ module.exports = {
         })
         .then(()=>{
           config.updateManifestVersion(module.name, module.version);
-          moduleController.restartModule(module.name);
+          //TODO: Need to revise where/when to restart an upgraded module based on background modules needing precedence to start first
+          // moduleController.restartModule(module.name);
```
tejohnso: If we don't start anything here, any new downloaded module versions will start up on the next restart. I think that's simplest and should be fine.

tejohnso: The whole background-foreground thing might be unnecessary too.  Modules know when they connect to LM. It should be reasonable for them (player) to hold off on sending whatever they need to send until after they've connected to LM.

PR #629, opened 2017-10-25
Revise install and upgrade flow

stulees:


stulees:
@tejohnso Added our discussed changes, cheers

PR #652, opened 2017-12-08
Use double backslash for player extraction on win

tejohnso:


tejohnso:
Yeah, it's rather unusual (at least to me) to have to double escape backslashes in a command line parameter.

PR #655, opened 2017-12-19
Deploy beta

stulees:
- Separation of beta and stable deploy tasks
- beta deploy does not push installer to root GCS and now updates beta manifests only
- stable deploy (triggered by `build/stable` branch) builds and pushes installer to root GCS, does not update any manifests
- Launcher now checks for BETA file bundled within `Installer/` directory, downloads remote beta manifest if exists, otherwise downloads remote stable manifest


.circleci/config.yml
```diff
@@ -54,6 +54,10 @@ jobs:
           root: .
           paths:
             - version-string
+      - run: |
+          if [ "${CIRCLE_BRANCH}" != "build/stable" ]; then
+            touch BETA
+          fi
       - run: mkdir -p ~/.ssh
```
tejohnso: I think we should be more cautious here and say:

```
touch BETA
if [ "${CIRCLE_BRANCH}" == "build/stable" ]; then
  rm BETA
fi
```

That way we default to BETA and if anything goes wrong with the comparison we don't harm STABLE.

.circleci/config.yml
```diff
@@ -230,17 +234,36 @@ jobs:
...
 
       - run: gsutil cp manifests/*.json gs://install-versions.risevision.com/master/
       - run: gsutil setmeta -h "Cache-Control:private, max-age=0" gs://install-versions.risevision.com/master/*.{sh,exe,json}
       - run: gsutil setmeta -h "Content-Disposition:attachment" gs://install-versions.risevision.com/master/*.sh
       - run: gsutil acl ch -u AllUsers:R gs://install-versions.risevision.com/master/*.{sh,exe,json}
+      - run: gsutil -m cp -p gs://install-versions.risevision.com/master/display-modules-beta-*.json gs://install-versions.risevision.com/
       - run: gsutil cp -p gs://install-versions.risevision.com/*.{exe,sh,json} gs://install-versions.risevision.com/backups/$(cat /tmp/shared-workspace/version-string/version)/
-      - run: gsutil -m cp -p gs://install-versions.risevision.com/master/* gs://install-versions.risevision.com/
       - run: gsutil -m cp -p gs://install-versions.risevision.com/master/* gs://install-versions.risevision.com/releases/$(cat /tmp/shared-workspace/version-string/version)/
```
tejohnso: Are the beta manifests going to be updated to include .../releases/... for the launcher module?
We should probably change this to .../releases/launcher... to be consistent with the others.

.circleci/config.yml
tejohnso: But I don't think we need it here.  We can check for ${CIRCLE_BRANCH} in build.sh

build.sh
```diff
@@ -47,6 +48,12 @@ do
   ./node_modules/.bin/electron-packager . "installer" --out="$ARCH_DIR/" --version="$ELECTRON_VERSION" --platform="$PLATFORM" --arch="$ELECTRON_ARCH" --ignore=".git|builds|^test" --asar=true --overwrite=true
   mv "$ARCH_DIR/installer-$PLATFORM-$ELECTRON_ARCH" "$VERSION_DIR/Installer"
 
+  # Conditionally bundle beta file
+  if [ -f "$BETA" ]
+  then
+    cp "$BETA" "$VERSION_DIR/Installer"
```
tejohnso: I think we can check the CIRCLE_BRANCH here instead of copying a file.  First `touch $VERSION_DIR/Installer/BETA` then if branch is stable/build `rm $VERSION_DIR/Installer/BETA`

.circleci/config.yml
stulees: @tejohnso Yes the beta manifests will have the `url` prop updated with ../releases/. Agreed, we should make consistent with other modules, will change that.

stulees:
@tejohnso Changes made, please review, thanks

package.json
```diff
@@ -6,7 +6,10 @@
   "author": "Tyler Johnson <mail@tyler-johnson.ca> (http://github.com/tejohnso)",
   "license": "ISC",
   "devDependencies": {
+    "electron": "1.7.8",
```
jungeunyoon: @stulees sneaky sneaky thanks for taking care of that 🥇

update-manifest-e2e.js
```diff
@@ -0,0 +1,41 @@
...
+  json.rolloutPct = (pct || pct === 0) ? pct : json.rolloutPct;
+  fs.writeFileSync(filePath, JSON.stringify(json, null, 2));
+} catch(e) {
+  return err(e);
+}
+
+function updateUrl(url, version, branch) {
+  let parts = url.split("/");
+
+  if (branch === "build/stable") {
```
tejohnso: If we're building stable we test staging/launcher/${version}/...

For non stable can't we say releases/launcher/${version}/.... ?  The difference would be more clear that way.

run-e2e-tests.sh
```diff
@@ -5,5 +5,23 @@ VERSION=$1
...
+if [ $ARCH == "lnx-armv7l" ]
+then
+  MANIFEST="$BASE_MANIFEST_NAME-lnx-arm.json"
+else
+  MANIFEST="$BASE_MANIFEST_NAME-$ARCH.json"
+fi
+
+mkdir -p e2e_$ARCH
+gsutil -m cp gs://install-versions.risevision.com/$MANIFEST e2e_$ARCH
+find e2e_$ARCH -name $MANIFEST -exec node update-manifest-e2e.js '{}' $BRANCH $VERSION 50 \;
```
tejohnso: Do we need to change the rollout%?  If so, why 50?

stulees: @tejohnso I followed the pct previously being applied. I'm not sure why it's 50 either, I'd assumed the reasoning involved launcher-e2e and didn't think it necessary to review. Do you recall why it has been 50?

PR #658, opened 2017-12-21
Expect beta manifests in beta subdur

tejohnso:


stulees:
@tejohnso Added a change, see last commit

PR #660, opened 2017-12-22
Treat player extraction differently

tejohnso:
Player build should eventually be changed to match other modules.


tejohnso:
@santiagonoguez It should be ok with latest commit

PR #662, opened 2017-12-29
Update rise-common-electron dependency version

jungeunyoon:


stulees:
@jungeunyoon Launcher is one module that should not have this updated. Even though logger.js is backwards compatible, lets leave it for now, thanks.

santiagonoguez:
nice to see you prevent us from destroying the universe !!! thanks !

PR #664, opened 2018-01-03
use echo for uninstall messages in PI

santiagonoguez:
Issue 17 in display control issue tracker. In order to avoid PI uninstaller failing on last message, this fix detects if we are in Raspbian and uses a simple echo notification if that's the case. Please review.


santiagonoguez:
does anyone has an idea why the UI integration test failed ?

stulees:
@santiagonoguez It seems to happen inconsistently. If you try re-running failed jobs, it will eventually pass.

PR #669, opened 2018-01-18
dont update executable on stable merge

santiagonoguez:
changed to only copy JSON files, and not executables.


.circleci/config.yml
```diff
@@ -264,7 +264,7 @@ jobs:
         - run: gsutil setmeta -h "Content-Disposition:attachment" gs://install-versions.risevision.com/master/*.sh
         - run: gsutil acl ch -u AllUsers:R gs://install-versions.risevision.com/master/*.{sh,exe,json}
         - run: gsutil cp -p gs://install-versions.risevision.com/*.{exe,sh,json} gs://install-versions.risevision.com/backups/$(cat /tmp/shared-workspace/version-string/version)/
-        - run: gsutil -m cp -p gs://install-versions.risevision.com/master/* gs://install-versions.risevision.com/
+        - run: gsutil -m cp -p gs://install-versions.risevision.com/master/*.json gs://install-versions.risevision.com/
```
stulees: @santiagonoguez I think it would be good to target manifests specifically via `display-modules-*.json` instead.

santiagonoguez: @stulees I changed the selection as you suggested ! now it's more specific.

PR #674, opened 2018-02-07
Add log "upgrading modules" to use for Upgrader reliability

stulees:
- Need to differentiate from a fresh install and focus module upgrade reliability only on actual module upgrades.
- Query that will be used for module upgrade stats - https://github.com/Rise-Vision/bigquery-queries/pull/27


stulees:
@tejohnso Our pre-modularized Upgrade reliability query didn't just take that into account, it was also conditional on specific events logged. I followed that same idea, however, you're right in that `"0.%"` is sufficient to ensure displays are not doing a fresh install. Please review my changes to the query here - https://github.com/Rise-Vision/bigquery-queries/pull/27 . I will close this PR pending your approval on query.

PR #676, opened 2018-02-21
Fix/download error workaround

santiagonoguez:
This intercepts the uncaught network error, and gives the installer some extra retries in order to complete. If the downloader component throws an error directly of the number of retries is surpassed the error will be propagated as part of the installer promise flow.

I tested using a staged installer with an slow connection and also manually interrupting network connection and seems to work well.

As a sideline comment, I noticed that some of the errors during disconnection are not reported to BQ because there's not network, but a local 'Could not log to bq' is reported.


installer/downloader.js
```diff
@@ -1,10 +1,74 @@
...
+}
+
+function isDownloading() {
+  return downloadContext.count > 0;
+}
+
+function retryDownload(error) {
+  if( downloadContext.count < MAX_DOWNLOAD_RETRIES ) {
+    const remoteModule = downloadContext.module;
+    log.all(`retrying download ${remoteModule.name}`, remoteModule.url, 20);
```
andrecardoso: What does 20 means here?

installer/downloader.js
```diff
@@ -1,10 +1,74 @@
...
+const url = require("url");
+const upgradeLimiter = requireRoot("installer/upgrade-limiter.js");
+const {platform, network} = require("rise-common-electron");
+const moduleCommon = require("common-display-module");
+const config = requireRoot("installer/config.js");
+const version = requireRoot("version.json");
+
+const MAX_DOWNLOAD_RETRIES = 4;
+
+let downloadContext = { count: 0 };
```
andrecardoso: Maybe add defaults to each property and rename it to `downloadState` to be more clear about what is stored in this object:

```javascript
let downloadState = { 
  count: 0, 
  promise: Promise.resolve(), 
  resolve: () => {}, 
  reject: () => {}, 
  module: null,
  savePath: null
}
```

installer/downloader.js
```diff
@@ -1,10 +1,74 @@
...
+  )
+  .catch(error =>
+    downloadContext.promise === currentPromise &&
+    finishDownload(downloadContext.reject, error)
+  );
+}
+
+function downloadFile(remoteModule, savePath) {
+  if (downloadContext.count > 0 ) {
+    throw new Error("A download is already in progress");
```
stulees: @santiagonoguez Can you clarify why we should throw an error here and not just `return` instead?

installer/downloader.js
santiagonoguez: @andrecardoso 20%, which is the percent modules show in the progress bar while downloading. I guess this was originally meant to be updating depending on the number of modules so we could see the bar moving ( before we made it indeterminate ). I know it's not nice, but I just kept the 20 used in other places of this section.

installer/downloader.js
santiagonoguez: @andrecardoso I like the idea, I'll update the code

installer/downloader.js
santiagonoguez: or were you asking why not ?
    return Promise.reject(new Error("A download is already in progress"));

installer/downloader.js
```diff
@@ -1,10 +1,74 @@
...
+function isDownloading() {
+  return downloadContext.count > 0;
+}
+
+function retryDownload(error) {
+  if( downloadContext.count < MAX_DOWNLOAD_RETRIES ) {
+    const remoteModule = downloadContext.module;
+    log.all(`retrying download ${remoteModule.name}`, remoteModule.url, 20);
+
+    tryDownload();
```
tejohnso: If there's a temporary issue with the network this is going to get called in rapid succession and the end result will be a very quick rejection. It might be better if it were wrapped in a `setTimeout` so the network has some time to possibly recover.

main.js
```diff
@@ -1,21 +1,37 @@
...
+const preventBQLog = process.env.RISE_PREVENT_BQ_LOG;
+const ui = requireRoot("installer/ui/controller.js");
+const ElectronProxyAgent = require("electron-proxy-agent");
+const downloader = requireRoot("installer/downloader.js");
+const externalLogger = requireRoot("installer/loggers/installer-logger.js");
+
+let displaySettings;
+
+function handleUncaught(prefix, error) {
+  if (downloader.isDownloading() && error.message && error.message.includes('net::ERR_')) {
```
tejohnso: I think we should be more specific here, for example `if (gotModuleThrewAnExceptionDueToIssueXXX()) {...}`.

And maybe the condition should check the error.stack for a specific line number. Generally we're strongly advised against attempting to recover from `uncaughtException` [1] so we should be really clear about the specific state we're attempting to recover from.

[1] https://nodejs.org/api/process.html#process_warning_using_uncaughtexception_correctly

> The correct use of 'uncaughtException' is to perform synchronous cleanup of allocated resources (e.g. file descriptors, handles, etc) before shutting down the process. It is not safe to resume normal operation after 'uncaughtException'.

installer/downloader.js
stulees: @santiagonoguez Yes wondering why halt halt execution by throwing an Error instead of just returning rejection.

installer/downloader.js
santiagonoguez: @andrecardoso added

installer/downloader.js
santiagonoguez: @andrecardoso changed

installer/downloader.js
santiagonoguez: @stulees changed it

old habits :D

santiagonoguez:
@tejohnso @stulees @andrecardoso I applied all the changes you suggested, as some timeout logic was added I added some tests and also did manual testing to ensure everything is working OK.

Please review again.

PR #679, opened 2018-02-23
dont wait for LMS connection

santiagonoguez:


installer/main-controller.js
```diff
@@ -54,7 +54,8 @@ function scheduleDiskSpaceCheck() {
 }
 
 function handleLMS() {
-  return moduleCommon.receiveMessages("installer").then((receiver)=>{
+  // don't wait on the promise receveMessages() return, or install process may hang
+  moduleCommon.receiveMessages("installer").then((receiver)=>{
     receiver.on("message", (message) => {
```
tejohnso: I don't think the comment makes it clear that an absence of a return function is what's being referred to.

With this diff, it's clear, but looking at the code later the comment won't be obvious. I think it should be removed or mention the lack of `return`.  Something like "No return here so that the install process won't hang if receiveMessages doesn't resolve."

santiagonoguez: @tejohnso changed it, thanks !

PR #680, opened 2018-02-26
Clear environment variables on startup

stulees:


stulees:
@santiagonoguez I want to validate with the staged installer this PR creates first as validation requires manually simulating a Viewer crash. Once validated I think this can be merged and appended to beta release in process. Thoughts?

santiagonoguez:
@stulees Sounds good to me

PR #681, opened 2018-02-27
Template for submitting an issue

alanclayton:
@tejohnso @santiagonoguez this is just a issue template to make sure we get all the information needed when a new defect is submitted.


docs/issue_template.md
```diff
@@ -0,0 +1,17 @@
+*NOTE: do not add Display ID or Company ID in this comment. If you have access, manually add them to the Trello card after submitting the defect. If SC is available, manually add a link to the Trello card after submitting the defect.*
+
```
tejohnso: @alanclayton Do we need to indicate where the Trello card is going to end up? Could even provide a link to the board.

alanclayton: @tejohnso good point. updated.

santiagonoguez:
@tejohnso for my part can be merged, @stulees do you have any problem with it ?

PR #682, opened 2018-02-27
Prefix "beta_" to installer version when logging if BETA installer

stulees:
- Using latest common-display-module to leverage common `isBetaLauncher()` function
- Prefixing `beta_` to version when logging to installer_events if it is a BETA installer


package.json
```diff
@@ -20,7 +20,7 @@
   "dependencies": {
     "angular": "1.5.9",
     "camelcase": "^2.1.0",
-    "common-display-module": "git://github.com/Rise-Vision/common-display-module.git#v2.1.4",
+    "common-display-module": "git://github.com/Rise-Vision/common-display-module.git#v3.0.1",
```
santiagonoguez: You're updating common-display-module from v2.1.4 to v.3.0.1. During that change messaging functions where extracted to their own module:

https://github.com/Rise-Vision/common-display-module/blob/v3.0.0/messaging.js

Launcher uses those in installer/main-controller.js  and it's associated unit test. So this line:

https://github.com/Rise-Vision/rise-launcher-electron/blob/master/installer/main-controller.js#L58

should not read from common, but from messaging. Same thing for it's related unit test.

If you have doubts regarding this issue contact me via hipchat.

stulees: @santiagonoguez Good catch, forgot about that. I'll make the change, cheers.

stulees:
@santiagonoguez Change made

PR #694, opened 2018-04-06
Try chromium61

tejohnso:


package.json
```diff
@@ -6,7 +6,7 @@
   "author": "Tyler Johnson <mail@tyler-johnson.ca> (http://github.com/tejohnso)",
   "license": "ISC",
   "devDependencies": {
-    "electron": "1.8.2",
+    "electron": "2.0.0-beta.7",
```
stulees: @tejohnso I didn't realize v2 is still in beta. Any concerns about us going stable with a beta Electron version?

santiagonoguez:
Also, this update will need to upgrade display control build also I guess, as happened last time ?

stulees:
@santiagonoguez Correct, DC needs to update serialport rebuild per arch/platform.

package.json
tejohnso: @stulees not ideal, but it's the 7th release and it's been in beta for over 40 days now. Plus we're releasing to our own BETA to see how it performs before going to stable.

PR #709, opened 2018-04-24
Append display id as query param to module download url

stulees:


installer/downloader.js
```diff
@@ -52,6 +53,10 @@ function downloadFile(remoteModule, savePath) {
     return Promise.reject(new Error("A download is already in progress"));
   }
 
+  const displaySettings = commonConfig.getDisplaySettingsSync();
```
tejohnso: Would prefer doing this sync read only once at startup / init time.
Maybe do something like [this](https://github.com/Rise-Vision/rise-launcher-electron/blob/master/main.js#L50)

stulees: @tejohnso Updated, see last commit, thanks

andrecardoso: Maybe we should cache the display settings value in common-display-module/config.js

PR #710, opened 2018-04-25
Use latest common-display-module for nld json in stable build

stulees:


stulees:
@santiagonoguez Added, see last commit

PR #713, opened 2018-05-02
increase manifest fetch retries

santiagonoguez:


tejohnso:
This is okay for unattended installs, but on initial download/install, it wouldn't be very nice to be sitting at an "upgrading" screen for so long. Especially since the user just downloaded and there shouldn't be anything to upgrade.

santiagonoguez:
@tejohnso what you recommend, maybe we can limit to 5 retries on initial downloads ?

tejohnso:
I think we should leave it at 2 for initial download / attended install.

stulees:
@santiagonoguez @tejohnso Agreed on leaving 2 for fresh attended install. Given that I think we could go 20 retries for upgrade/unattended.

santiagonoguez:
@tejohnso @stulees now I differentiate between attended and unattended, please review again

PR #727, opened 2018-05-25
Fix/service worker dont register

santiagonoguez:
Please let me know if you agree with merging this fix to beta, according to my comments in the Trello card. I won't merge until everyone is OK with it.


andrecardoso:
@santiagonoguez can we close this?

PR #734, opened 2018-06-07
Feature/install excludes display control module

santiagonoguez:
I've tested it manually also.


installer/display-control-feature.js
```diff
@@ -0,0 +1,79 @@
+const network = require("rise-common-electron/network");
+
+const apiEndpoint ="https://www.googleapis.com/storage/v1";
+const bucketName = "risevision-display-notifications";
+const urlParams = "alt=media&ifGenerationNotMatch=-1";
```
tejohnso: I think `ifGenerationNotMatch=-1` can be removed.

installer/display-control-feature.js
```diff
@@ -0,0 +1,79 @@
...
+const urlTemplate = `${apiEndpoint}/b/${bucketName}/o/FILEPATH?${urlParams}`;
+const maxRetries = 5;
+const oneSecond = 1000;
+const downloadPercent = 20;
+
+function defaultDelay() {
+  return new Promise(resolve => setTimeout(resolve, oneSecond));
+}
+
+function isDisplayControlDownloadRequired(displayId, manifest, delay = defaultDelay) {
```
tejohnso: Would be good to keep the variable consistent as *localManifest* so as not to be confused with remote manifest.

installer/display-control-feature.js
```diff
@@ -0,0 +1,79 @@
...
+
+  return delay()
+  .then(() => checkIfDisplayControlIsEnabledInConfigurationFile(url, retries - 1, delay));
+}
+
+function isDisplayControlEnabled(content) {
+  const interfaceLine = (content || '').split("\n")
+  .find(line => line.startsWith('interface='));
+
+  return interfaceLine && ! interfaceLine.trim().endsWith('=');
```
tejohnso: Please remove the extra space after the `!` :)

installer/display-control-feature.js
```diff
@@ -0,0 +1,79 @@
...
+    if (error.stack && error.stack.includes('Response code 401')) {
+      return false;
+    }
+
+    return retryCheck(url, retries, `${url} | ${error.stack}`, delay);
+  });
+}
+
+function retryCheck(url, retries, errorMessage, delay) {
+  if (retries == 0) {
```
tejohnso: `<=` 0 would be safer here.

installer/downloader.js
```diff
@@ -115,21 +116,36 @@ network.registerObserver((stats)=>{
   log.progress("downloading latest version", (totalReceived / totalExpected * 100));
 });
 
+function getModulesToBeDownloaded(displayId, remoteData) {
+  const localManifest = moduleCommon.getManifest();
+
+  return displayControl.isDisplayControlDownloadRequired(displayId, localManifest)
```
tejohnso: I'd like an extra `.catch(()=>false)` here just to be sure the display control check can't disrupt.

installer/display-control-feature.js
```diff
@@ -0,0 +1,79 @@
...
+const oneSecond = 1000;
+const downloadPercent = 20;
+
+function defaultDelay() {
+  return new Promise(resolve => setTimeout(resolve, oneSecond));
+}
+
+function isDisplayControlDownloadRequired(displayId, manifest, delay = defaultDelay) {
+  if (manifest['display-control']) {
+    return Promise.resolve(true);
```
tejohnso: Okay, so once a display has the display control module, it's going to always upgrade it as per *rolloutPct* even if it is later disabled for that display. We might want to revisit that later, as it could add bandwidth overhead for nothing. But I think it's safer this way for now.

installer/display-control-feature.js
```diff
@@ -0,0 +1,79 @@
+const network = require("rise-common-electron/network");
+
+const apiEndpoint ="https://www.googleapis.com/storage/v1";
+const bucketName = "risevision-display-notifications";
+const urlParams = "alt=media";
+const urlTemplate = `${apiEndpoint}/b/${bucketName}/o/FILEPATH?${urlParams}`;
+const maxRetries = 5;
+const oneSecond = 1000;
+const downloadPercent = 20;
```
andrecardoso: This variable is only used on the log calls. Why are we sending this value to logs?

installer/downloader.js
andrecardoso: Isn't it better to add this `catch(() => false)` to `isDisplayControlDownloadRequired` implementation?

installer/downloader.js
```diff
@@ -115,21 +116,37 @@ network.registerObserver((stats)=>{
...
+  .catch(() => false)
+  .then(displayControlDownloadRequired => {
+    return remoteData.modules
+    .filter(remoteModule => remoteModule.name && remoteModule.version && remoteModule.url)
+    .filter(remoteModule => {
+      const moduleName = remoteModule.name;
+      const localModule = localManifest[moduleName];
+      const { percentageWasMet } = remoteData;
```
andrecardoso: I think we can return early an empty list if `percentageWasMet` is false and avoid the network calls to check if display control is required.

installer/downloader.js
```diff
@@ -115,21 +116,37 @@ network.registerObserver((stats)=>{
...
+
+      if (!localModule) {
+        return moduleName === 'display-control' ?
```
andrecardoso: Are we going to install display-control regardless of `percentageWasMet` value?

installer/downloader.js
```diff
@@ -115,21 +116,37 @@ network.registerObserver((stats)=>{
   log.progress("downloading latest version", (totalReceived / totalExpected * 100));
 });
 
+function getModulesToBeDownloaded(displayId, remoteData) {
```
andrecardoso: This class already has the displayId stored on `displaySettings`. Maybe you can use it instead of adding as parameter. That would avoid changing the other methods: `tryUpgrade`, `installUnattended`, etc

installer/display-control-feature.js
santiagonoguez: @andrecardoso it's a parameter for the log call that used to control the progress bar. As progress bar is now undetermined maybe we should revise all these calls to get rid of these parameter in all the project.

installer/downloader.js
santiagonoguez: @andrecardoso either way is fine, but I like it here because it reinforces the idea that it's an extra protection against errors in that implementation

installer/downloader.js
santiagonoguez: @andrecardoso we'll be installing display control even if percentageWasMet is false, if it's required

installer/downloader.js
santiagonoguez: @andrecardoso yes, because if it's required and it does not exist, it's a new module and should be installed

installer/downloader.js
santiagonoguez: @andrecardoso I saw that, but I did not know if I could trust it was already stored in all cases. But I'll double check and use it if it's always there for both scenarios ( attended and unattended ).

santiagonoguez: @andrecardoso  it's being set here: https://github.com/Rise-Vision/rise-launcher-electron/blob/master/main.js#L52

so as you said, it's always available, I'll make the change to use that, thanks !

santiagonoguez: @andrecardoso  though, now I wonder, it doesn't get updated if later in the process user assigns or changes the display id... maybe passing the parameter is safer, even for the code that's using that displaySettings variable, what do you think ?

installer/display-control-feature.js
andrecardoso: sounds good

installer/downloader.js
andrecardoso: Yeah.. It is just that returning false in case of an error is part of the answer to the "is display control required" question that the method is responsible for

installer/downloader.js
andrecardoso: Keeping it as a parameter would not fix this issue because you're reading the value from the same displaySettings object: https://github.com/Rise-Vision/rise-launcher-electron/pull/734/files#diff-050cb7aee8e5c91d93f86364a9eb5a93R144

I don't like having copies of `displaySettings` spread through the code. I think we're doing it for caching purposes because the `getDisplaySettingsSync` method reads the file every time. I'd rather keep the cached object in config and make it watch the ini file for changes to update the cached object.

andrecardoso: I'm not suggesting you to implement this in this PR :-)

santiagonoguez: @andrecardoso or change it only through functions in config.js, so you don't have to watch it assuming you won't be changing it while it's running. Last I saw launcher is the only one that writes to that file. I'll take a deepder look into these considerations.

andrecardoso: @santiagonoguez makes sense. Thanks

santiagonoguez:
@andrecardoso based on our conversation I think the safest here is to read the display id from the settings INI file. If we later cache the value only on common.js, this will use the latest value without having to reload the file each time. Please review again.

PR #742, opened 2018-06-18
Fix/centralize display settings

santiagonoguez:
This is the final fix, with the changes discussed this morning during Standup. I've also tested it manually.


installer/main-controller.js
```diff
@@ -163,11 +163,17 @@ function ready() {
   });
 }
 
+function logInstallComplete() {
+  const displayId = config.getInstallOrUpgradeDisplayId();
+
+  log.all("install complete", displayId, 100);
```
andrecardoso: Is this going to log the temporary display id on new installs?

santiagonoguez: @andrecardoso yes, that's why the function was called getInstallOrUpgradeDisplayId, and not just getDisplayId

andrecardoso: I think in this case we want to log the display id in the event_details as we're already logging the temporary display id in the display_id field

santiagonoguez: @andrecardoso No, with this PR the display_id field no longer contains the temporary display id, but the definitive one. That's why I'm adding the temporary display id on event_details for new installs and the reliability queries will need to read it from here: https://github.com/Rise-Vision/bigquery-queries/pull/59

andrecardoso: @santiagonoguez sounds good. Thanks

santiagonoguez:
@andrecardoso @rudibravo switched to option A, also tested it manually, please review again

PR #744, opened 2018-06-22
wait for promise to be resolved before continuing

santiagonoguez:
While reviewing code looking for No path/version found for local-messaging possible causes I realized we were not correctly waiting for this promise to be completed before continuing, so this could cause some permission related failure at some point.


andrecardoso:
Good catch!

andrecardoso:
[(2) ChrOS: support reboot](https://trello.com/c/LRnRpXeL/4100-2-chros-support-reboot)

Issue #746, opened 2018-06-27
Links to install Rise Player Linux and Raspbian shell scripts open in new tab instead of prompting to download.


btfrise:
- Operating System Version: Mac OS X 10.13.5
- Chrome Version: 67.0.3396.99 (Official Build) (64-bit)

##### What is the expected behavior?
Clicking a link to download a Linux or Raspbian installer should prompt for download.

##### What is the actual behavior?
Clicking a link to download a Linux or Raspbian installer opens a new tab with the contents of the script displayed.

##### What steps can be taken to replicate the problem? 

1. In Chrome go [here](https://help.risevision.com/hc/en-us/articles/115001872283-Install-Rise-Player-on-Linux) or create a new Display.
2. Click "Download Player for ..."

##### Notes
Cannot replicate in Firefox, or with other .sh files around the web. Test by clicking the ticketbleed.sh hyperlink [here](https://testssl.sh/).

btfrise:
![](https://github.trello.services/images/mini-trello-icon.png) [(1) Rise Player downloads for Linux and Raspbian open in new tab instead of prompting to download in Chrome.](https://trello.com/c/385ejwxr/8221-1-rise-player-downloads-for-linux-and-raspbian-open-in-new-tab-instead-of-prompting-to-download-in-chrome)

btfrise:
Andre fixed this but requested I leave open.

rudibravo:
@btfrise we fixed the issue with today's release. Can you please verify and close? Thanks

PR #758, opened 2018-08-23
Update executables on stable build

rudibravo:
Fix for issue #746. We tried not to publish the executables right away, but it didn't work out and we are copying the files manually after each release.
Reverted the changes and publishing the files again.
@Rise-Vision/delivery please review


rudibravo:
[(1) Defect 746: Links to install Rise Player Linux and Raspbian shell scripts open in new tab instead of prompting to download.](https://trello.com/c/GS2U2xxt/4470-1-defect-746-links-to-install-rise-player-linux-and-raspbian-shell-scripts-open-in-new-tab-instead-of-prompting-to-download)

PR #774, opened 2018-11-14
Clear proxy config on ini file

andrecardoso:


andrecardoso:
[(2) Defect 763: Proxy settings not being cleared when proxy system setting is removed](https://trello.com/c/HsJm3Tfb/4694-2-defect-763-proxy-settings-not-being-cleared-when-proxy-system-setting-is-removed)

PR #775, opened 2018-11-14
Remove duplicated network check

andrecardoso:


andrecardoso:
[(1) Electron: Remove duplicated network check from launcher](https://trello.com/c/54cjCPxC/4714-1-electron-remove-duplicated-network-check-from-launcher)

Issue #816, opened 2019-09-17
Rise Player on Win 10 sometimes doesn't automatically start


PatrickA-RiseVision:
**Describe the bug**
Rise Player doesn't start automatically.  After reinstalling or manually placing the lnk file into the startup folder, it will start automatically for a few days, but at some point the file disappears from startup.

When the issue is happening, this error is logged when Rise Player is started manually: [enoent.txt](https://github.com/Rise-Vision/rise-launcher-electron/files/3623347/enoent.txt)

We verified the install path is correct on the affected machine, but the Rise Vision Player.lnk file it's trying to rename doesn't exist on either a working machine or the affected machine.

**Reach**
This is currently reaching several displays for a single customer.  Not all players are affected at the same time, but they've had at least 20 of them affected at one time.

**Impact**
There is a workaround, which is to reinstall player from scratch or manually place the lnk file in the startup directory.  It will work for a seemingly unpredictable amount of time, and suddenly fail again.

**To Reproduce**
I haven't been able to reproduce this behavior on my player.  We can coordinate with the customer to work with an affected player as needed.  I'll provide additional details in the card.

**Expected behavior**
Rise Player starts automatically when the Media Player is started.

**Screenshots**
n/a

**Desktop (please complete the following information):**
 - OS: Windows 10 Education

**Additional context**
- These affected players are on a domain.  It may be related to the issue, but we're not sure.

olegrise:
Couple of things here:
1) It seems that the root of the problem is that Player autostart shortcut is being created under [`process.env.APPDATA](https://github.com/Rise-Vision/rise-common-electron/blob/0e1c39d350eefb047b73176d3cb519fc0a34c245/platform.js#L76) folder. The problem with `process.env.APPDATA` is that it points to "C:\\Users\\username\\AppData\\**Roaming**\\". The roaming folder can move with user profile from PC to PC ([more details](https://support.pointlogic.com/faq/troubleshooting/accessing-the-appdata-folder)). The alternative is to use `process.env.ALLUSERSPROFILE` which will install shortcut for all users on that machine.
2) There is a chance that the shortcut is being removed by Windows Maintenance ([see here](https://www.ghacks.net/2010/03/30/fix-windows-7-desktop-shortcuts-disappearing/)). It should be an easy task to turn off Windows Maintenance and see if that help.

@tejohnso any thoughts on the above?

PatrickA-RiseVision:
[...]

PR #821, opened 2019-10-02
Fix issue 816

olegrise:
## Description
Fix for #816 

## Motivation and Context
The Player installer recreates auto-startup shortcut every time it runs. In some cases it deletes existing shortcut, but then it fails to create a new one. The fix ensures that existing shortcut is not deleted if creation a new shortcut fails.

## How Has This Been Tested?
Manually and with unit tests.

## Release Plan:
- As the Submitter, upon requesting review of this pull request, I confirm that the [Release Checklist](https://help.risevision.com/hc/en-us/articles/360031119991) has been completed. 
- As the Reviewer, upon approving the changes in this PR, I confirm I have reviewed and I agree that the [Release Checklist](https://help.risevision.com/hc/en-us/articles/360031119991) has been completed

#### Release Checklist Items Skipped?
If any Release Checklist items were intentionally skipped, please provide which ones and the reasons why


installer/autostart/autostart.js
```diff
@@ -53,10 +53,16 @@ module.exports = {
...
-      return platform.deleteRecursively(module.exports.autostartPath());
+      if (platform.fileExists(shortCutPathTemp)) {
+        log.debug("deleting existing autostart");
+        return platform.deleteRecursively(module.exports.autostartPath());
+      }
     })
     .then(()=>{
-      return platform.renameFile(shortCutPathTemp, module.exports.autostartPath());
+      if (platform.fileExists(shortCutPathTemp) && !platform.fileExists(module.exports.autostartPath())) {
+        log.debug("copying temp shotrcut file to autostart folder");
```
andrecardoso: Typo here: `shortcut`

installer/autostart/autostart.js
```diff
@@ -53,10 +53,16 @@ module.exports = {
       return config.createWindowsShortcut(version, shortCutPathTemp, backgroundScriptPath, "start.bat --unattended");
     })
     .then(()=>{
-      return platform.deleteRecursively(module.exports.autostartPath());
+      if (platform.fileExists(shortCutPathTemp)) {
+        log.debug("deleting existing autostart");
```
andrecardoso: Shouldn't we log to `all` or `external` so that we can query it remotely?

installer/autostart/autostart.js
```diff
@@ -53,10 +53,16 @@ module.exports = {
...
+        return platform.deleteRecursively(module.exports.autostartPath());
+      }
     })
     .then(()=>{
-      return platform.renameFile(shortCutPathTemp, module.exports.autostartPath());
+      if (platform.fileExists(shortCutPathTemp) && !platform.fileExists(module.exports.autostartPath())) {
+        log.debug("copying temp shotrcut file to autostart folder");
+        return platform.renameFile(shortCutPathTemp, module.exports.autostartPath());
+      }
```
tejohnso: Why is this a delete then rename instead of just a rename (which will overwrite).

olegrise: @tejohnso good question. That's how it was before and I didn't touch it because maybe there is a reason for that.

tejohnso: This leaves another potential issue where the delete worked, but the rename fails.

I think it would be safer to leave the delete out and try a rename.

It's possible that older versions of Electron would throw an error instead of overwriting. So I think it should be tested with the overwrite and if it works, drop the delete.

olegrise: @tejohnso I removed the delete operation and tested. Overwrite works. See the latest commit. Thanks.

